PR #3, opened 2015-06-26
On Greg's advice, ripping out the (channel, headline) tuple and just in-lining the headline to a story

tvaughan77:


willvanwazer:
Is there a possibility of revisiting this decision? I was about to make a pull request to re-instutite the headline per channel concept, as we need different headlines for the channels web, mobile and social to match the current functionality of WaPo.com.

tvaughan77:
Yes, we're going to need to revisit this.  Greg is setting up a pow-wow with some Content SMEs and we'll bring it up.

PR #14, opened 2015-08-28
Renaming Story to Collection, and then adding new subtypes of collection.

willvanwazer:
These are called Story and Gallery, which are entirely the same except for presentational differences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,12 @@ The latest version of the ANS schema is the 0.3.0-SNAPSHOT release, which includ
 ## Schema files
 ANS Schema files are defined with the [JSON Schema specification](https://spacetelescope.github.io/understanding-json-schema/index.html).  Schemas are defined in individual files under the [src/main/resrouces/schema/ans/_version_/](src/main/resources/schema/ans/v0_3/) directory. 
 
-Small, self-contained, attributes such as "had copyright" or "had unique id" are modeled as **traits**, which are combined together or applied directly to more complicated schema objects for the purpose of re-usability.  The "content.json" schema, for example, composes the traits "dated", "credited", "locale", "located", and "copyrighted".  The interesting schemas (story.json, image.json, and video.json) include/descend from that basic content schema and provide additional schema elements as appropriate.
+Small, self-contained, attributes such as "had copyright" or "had unique id" are modeled as **traits**, which are combined together or applied directly to more complicated schema objects for the purpose of re-usability.  The "content.json" schema, for example, composes the traits "dated", "credited", "locale", "located", and "copyrighted".  The interesting schemas (collection.json, image.json, and video.json) include/descend from that basic content schema and provide additional schema elements as appropriate.
 
 One can use a variety of third party tools to validate their content against the schema files hosted here on github publicly.
 
 ## Schema Overview
-The ARC Schema is rooted at an "ANS" node that defines an ID and Type field; the ID should be a globally unique identifier to the concrete/subclass instance of content and the Type field should describe to higher-level languages what type of content the concrete instance is (i.e. "story" or "image")
+The ARC Schema is rooted at an "ANS" node that defines an ID and Type field; the ID should be a globally unique identifier to the concrete/subclass instance of content and the Type field should describe to higher-level languages what type of content the concrete instance is (i.e. "collection" or "image")
 
 The hierarchy of content is diagramed here: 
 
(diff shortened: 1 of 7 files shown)


src/test/java/com/washingtonpost/legacy/README.md
```diff
@@ -1,4 +1,4 @@
-# Examples of current story content
-This directory holds examples of all the content that goes in to a current WashingtonPost.com story or video
+# Examples of current collection content
```
tvaughan77: Can you roll back your find & replace for this file?  It's talking about legacy story (small 's', the concept, not big 'S', the class)

src/test/java/com/washingtonpost/arc/ans/v0_3/model/TestStory.java
```diff
@@ -7,10 +7,10 @@
 import org.junit.Test;
 
 /**
- * <p>Tests that JSON we expect to be a valid "Story" data file serializes correctly and validates
+ * <p>Tests that JSON we expect to be a valid "Collection" data file serializes correctly and validates
  * against the JSON schema</p>
  */
-public class TestStory extends AbstractANSTest<Story> {
+public class TestStory extends AbstractANSTest<Collection> {
```
tvaughan77: This class should be TestCollection

src/test/java/com/washingtonpost/arc/ans/v0_3/model/TestStory.java
```diff
@@ -19,13 +19,13 @@ String getSchemaName() {
...
     Class getTargetClass() {
-        return Story.class;
+        return Collection.class;
     }
 
     @Test
     public void testStoryGood() throws Exception {
         testJsonValidation("story-fixture-good", true);
-        Story story = testClassSerialization("story-fixture-good");
+        Collection story = testClassSerialization("story-fixture-good");
```
tvaughan77: variable name should be 'collection' and fixture name should be 'collection-fixture-good'

src/test/java/com/washingtonpost/arc/ans/v0_2/model/TestStory.java
```diff
@@ -8,7 +8,7 @@
 import org.junit.Test;
 
 /**
- * <p>Tests that JSON we expect to be a valid "Story" data file serializes correctly and validates
+ * <p>Tests that JSON we expect to be a valid "Collection" data file serializes correctly and validates
```
tvaughan77: roll this find & replace back... you shouldn't change anything under the ans/v0_2 directory because it's already released

src/main/java/com/washingtonpost/arc/ans/v0_3/model/ANS.java
```diff
@@ -22,7 +22,7 @@
     @Type(value = Media.class, name = "media"),
     @Type(value = Image.class, name = "image"),
     @Type(value = Video.class, name = "video"),
-    @Type(value = Story.class, name = "story"),
+    @Type(value = Collection.class, name = "story"),
```
tvaughan77: I don't think this is right.  We want to instantiate an actual Story.class when a JSON block declares itself of type "story"

Likewise, you should add a new Type annotation that maps "Gallery.class" to the name="gallery"

README.md
```diff
@@ -18,12 +18,12 @@ The latest version of the ANS schema is the 0.3.0-SNAPSHOT release, which includ
...
 
 One can use a variety of third party tools to validate their content against the schema files hosted here on github publicly.
 
 ## Schema Overview
-The ARC Schema is rooted at an "ANS" node that defines an ID and Type field; the ID should be a globally unique identifier to the concrete/subclass instance of content and the Type field should describe to higher-level languages what type of content the concrete instance is (i.e. "story" or "image")
+The ARC Schema is rooted at an "ANS" node that defines an ID and Type field; the ID should be a globally unique identifier to the concrete/subclass instance of content and the Type field should describe to higher-level languages what type of content the concrete instance is (i.e. "collection" or "image")
```
tvaughan77: I'd roll this example back to using "story" because (a) it's a real type and (b) it's more immediately obvious to the reader of a schema overview what a "story" is compared to a "collection"

README.md
```diff
@@ -32,7 +32,7 @@ The hierarchy of content is diagramed here:
 ## Examples
 Two examples are provided of content documents that validate against the Washington Post ANS schema:
 
-1. [An Example Story](src/test/resources/com/washingtonpost/arc/ans/v0_3/model/story-fixture-tiny-house.json)
+1. [An Example Story](src/test/resources/com/washingtonpost/arc/ans/v0_3/model/collection-fixture-tiny-house.json)
```
tvaughan77: Did you actually change the name of the file?  I don't think you did, so maybe just keep it as "story-fixture-tiny-house.json"?

src/main/java/com/washingtonpost/arc/ans/v0_3/model/ANS.java
tvaughan77: And I suppose it's sort of a TBD if we would want someone to declare a block of json as a type="collection" ... we do allow a type="media" which is kind of the parallel concept over in the "video & image" side of the schema.  So if you think you might want a concrete Collection instance at some point, go ahead and add a Type annotation that maps Collection.class to name "collection"

PR #22, opened 2015-11-02
Adds unordered and ordered lists, blockquote

gengel:
Should be self-explanatory?


tvaughan77:
At a technical level, this looks good with the exception that I think we need to add BQ, OL, UL to the list of allowed content elements in content_element.json.  Try adding an example of each new type to one of the story-good-*.json fixtures and see if the tests still pass to see what I mean.  

Also (at a technical level), why is an unordered_list an array of list_items whereas an ordered_list is an array of list_elements?  It seems like there's one-too-many objects in there?

Okay, so popping up to the sort of strategic/design level -- and understanding that this isn't a 1:1 fit with what Will & Matt have already done -- it seems like the ListItem in this PR is really the same thing as a ContentElement in the existing schema and that OL and UL could be replaced by "List, with an ordering = {ordered, unordered, null}".  Armed with such a List we could replace the Collection.ContentElements with a Collection.List (ordering = null)

gengel:
The ListItem in UnorderedList is wrong, that should have been ListElement (which is just an abstraction to allow nesting lists)

The "content elements" in a Story (or whatever) are not a semantic "list" in the HTML sense.  To my mind, the main beneficiaries of the schema are the Editing UI and various renderers. If so, the renderer should not just look for a "list" as a top-level Story property and then start rendering <li> elements for each paragraph, video, gallery, etc.  I suppose we could set up the editor/renderer to look for a "list" object and then not render anything special for "ordering=null" but then, what is the point?  When push comes to shove I just don't think a "ul/ol" is semantically the same as the sequence of content elements that make up an article (or gallery) body.

However, if we're not bound the Matt/Will's schema, then there are certainly narrower changes that could be made.  I'm fine combining UL/OL into one "list" type with an "ordered" boolean.  And right now ListItem is essentially just a Text element, so we could  make ListElement include Text and drop ListItem ... OR we could go the broader route of allowing a broader subset of content elements to also be list elements. The latter would have serious repercussions for the editor, though, I would think.  

Maybe the most forwards-compatible option might be just say ListItem is "oneof" and then only include Text and Image for now, opening it up further if people really need other things.

tvaughan77:
> When push comes to shove I just don't think a "ul/ol" is semantically the same as the sequence of content elements that make up an article (or gallery) body.

Fair 'nuff...

>  And right now ListItem is essentially just a Text element, so we could make ListElement include Text and drop ListItem

Yeah, that sounds good/better.  We're still in -SNAPSHOT, so let's make these small changes, ship it, and see how it goes

PR #26, opened 2015-11-10
Audio element

gengel:


src/main/resources/schema/ans/v0_3/audio.json
```diff
@@ -0,0 +1,38 @@
...
+                "loop": {
+                    "description": "Whether looping is enabled.",
+                    "type": "boolean"
+                },
+                "preload": {
+                    "description": "Whether preload is enabled.",
+                    "type": "boolean"
+                }
+            },
+            "required": ["source"]
```
tvaughan77: why is source required?

tvaughan77: oh ... i see ... is source maybe better named "url" ?  or "sourceUrl" ?

PR #28, opened 2015-11-20
Adds "channel" trait

gengel:
Adds "channel" trait in JSON Schema and makes all possible content elements descend from ContentElement in Java model

### DIFF
--- a/src/test/resources/com/washingtonpost/arc/ans/v0_3/model/text-fixture-good-channels.json
+++ b/src/test/resources/com/washingtonpost/arc/ans/v0_3/model/text-fixture-good-channels.json
@@ -1,0 +1,6 @@
+{
+    "_id": "0987654321",
+    "type": "text",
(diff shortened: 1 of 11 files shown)


src/test/resources/com/washingtonpost/arc/ans/v0_3/model/text-fixture-good-channels.json
```diff
@@ -0,0 +1,6 @@
+{
+    "_id": "0987654321",
```
tvaughan77: can we add some channels to content_elements in the main story-fixture-good JSON?  that's the JSON we link to from the README.md and that Greg F often asks for as an example of what to use/show?

tvaughan77: otherwise, as long as this "mvn clean install"s this looks good to merge; fire away

PR #31, opened 2015-11-30
Changing headline to be an object for ease of use.

willvanwazer:
Rather than having headlines have the structure of ```List<Headline>```, it makes more sense to have headlines take the format of ```Map<String, String>```. This makes code to access a particular headline in JSPs as simple as ```${story.headlines.twitter}```, rather than requiring a developer to iterate through all the headline objects to find the one of the correct type.

This also discourages people from beginning to use the convention of a particular type of headline being in a specific numbered spot, like "the twitter headline is always headline 2 in the list, so just use ```headlines.get(1)``` to grab it. Coincidentally, we actually used this approach in our own testing for this module, which illustrates the issues with this approach--adding another headline to our test objects would have totally broken our code, which is bad.


gengel:
While we're looking at this, how do you feel about requiring the `default` headline? Eric points out that losing the list format means there's no clear precedence for which headline to fallback to unless certain keys are defined.  This would make serialization a little harder - probably a Headlines class again, with just a default and a Map<String,String>...  Alternatively, we could (1) require the default field in the JSON schema, leave Map<String,String> as is and allow the implementation to assume that clients are smart enough to add a "default" key on their own, or (2) make the `default` headline a convention rather than definition.

PR #40, opened 2016-01-28
Schema refactor

gengel:
- Adds test coverage 
- Changes JSON validator to one that "really works"
- Updates fixtures to comply with tests
- Fixes minor bugs in schema definitions revealed by tests


rgfindl:
I think it would be good to show some failing story tests.

PR #61, opened 2016-06-22
0 5 6

rgfindl:
http://confluence.washpost.com/display/ETT/0.5.6


lib/transforms.js
```diff
@@ -493,9 +493,45 @@ var versions = {
...
+              });
 
+              result[key] = sites;
+            }
+          } else if ( key === "owner") {
+            // Update owner.  Add 'sponsored' attribute.
+            value.sponsored = false;
+            result[key] = value;
+          } else {
+            result[key] = value;
```
gengel: L527 - This should be `if (key !== "additional_properties") { result[key] = convert(value) }`, I think, since ans objects can be nested

gengel: EDIT: Probably something more like `if (key === 'additional_properties') { result[key] = value; } else { result[key] = convert(value); }` so as not to delete additional_properties

PR #65, opened 2016-07-01
CA 47

gengel:
* Removes dead-end upverters for 0.4.5 -> 0.4.6, 0.4.6 -> 0.4.7
* Adds internal version sequence doc
* Moves upverters into their own files, dynamically imported based on version sequence
* Adds command line interface
* Adds "create" command to generate schemas, upverter and test from current version
* Adds "cleanup" command to delete above
* getSchemas and getTransforms now pass "null" to callback on success (more idiomatic nodejs)
* package version de-coupled from ANS version


lib/util.js
```diff
@@ -1,16 +1,6 @@
 'use strict';
```
rgfindl: Is this file still needed?

gengel: Nope, I'll remove it

rgfindl:
Looks awesome.  Much cleaner.  Good work.  Merge when ready.

PR #88, opened 2017-01-06
Tags

### LINKED ISSUE #87: Proposal: improve support for tags, add categories
**Problem**
ANS Tags do not support wordpress categories

**Proposed Solution**
Add to utils/tag.json:
 - slug = tag or category slug, e.g. 'my-category-slug'
 - text = tag or category name ~~(currently listed in spec as "tag")~~
 - description = tag or category description
 - additional_properties = (same as elsewhere, used for non-standard metadata)

**Example**
```tags: [
{
_id: "DGBTWSJS2BHNTPWLVVUYPM6T6E",
  text: "My Category Name",
  additional_properties: {wordpress_type: "category"},
  slug:  "my-category-name",
  description: "My superfab category"
},
{
_id: "AMKDOSHIJ5FYVJS5UASQZ2G2CY",
text: "My Tag Name",
  additional_properties: {wordpress_type: "tag"},
slug: "my-tag-name",
description: "My unparalleled tag"
}
]
```

**Question**
Currently, tags have two properties, on the api output: _id and "tag".  Am I misreading the spec when it appears that the properties should be _id and "text"?

gengel:
Fixes #87

### DIFF
--- a/lib/transform_utils.js
+++ b/lib/transform_utils.js
@@ -2,7 +2,8 @@
 
 var _ = require('lodash');
 
-var top_level_types = ["story", "video", "image", "gallery", "revision"];
+var top_level_types = ["story", "video", "image", "gallery", "results", "audio", "redirect"];
+var recursion_keys = [ 'promo_items', 'related_content', 'content_elements' ];
 
 // Given a version number, generates a function that sets input.version to that version number
 var version_incrementer = function(new_version) {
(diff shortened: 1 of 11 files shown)


lib/transform_utils.js
```diff
@@ -2,7 +2,8 @@
 
 var _ = require('lodash');
 
-var top_level_types = ["story", "video", "image", "gallery", "revision"];
+var top_level_types = ["story", "video", "image", "gallery", "results", "audio", "redirect"];
```
chadmoone: Should we have lost `revision` here?  Tests pass, so maybe it was in erroneously before, but I didn't see mention of it elsewhere here or in the proposal.

gengel: Yeah, it was erroneous -- should have been "results"

PR #109, opened 2017-03-10
Allow null edition_time in publication pitch event

jpblair:
I wasn't able to verify that the tests pass since they appear to be broken for an unrelated reason.


gengel:
Closing this for now.

PR #131, opened 2017-04-20
Adds proposals for chart, statistic series elements.

gengel:
This pull request adds proposals for two new content element types: `charts` and `statistic_series`.  

I am not wedded to the names, but the basic reasoning for the two types is as follows:

- Wanted to model some common current newsroom usage of "tables" (here meaning Methode/HTML table support)
- Wanted to target elements as specifically as possible
- Wanted to represent the data in an implementation-agnostic way (nothing specific to HTML, Methode, WordPress, etc)
- At the same time, needs to have a clear rendering solution in HTML, native apps, Ellipsis, etc.

Examples of the types of data these two elements represent are given in their respective proposal documents, which should be considered together.

This proposal should be taken to preclude adding additional element types in the future for different use cases (e.g., bar graphs)

### DIFF
--- a/docs/proposals/2017-04-20 - Charts.md
+++ b/docs/proposals/2017-04-20 - Charts.md
@@ -1,0 +1,201 @@
+# Problem
+
+Talking to publishing support, it seems as though the most often use case for tables was when doing a "basic data chart."  Estimated usage is 1-2 times a week throughout the entire newsroom. The following are some recent examples:
+
+https://www.washingtonpost.com/lifestyle/kidspost/who-deserves-a-shot-at-the-nba-all-star-game/2017/01/25/c9995306-de5e-11e6-918c-99ede3c8cafa_story.html
+
+https://www.washingtonpost.com/news/capital-weather-gang/wp/2017/03/02/does-a-record-warm-winter-mean-a-super-hot-summer-in-d-c/
+
+Ellipsis needs to create data charts like these within an article. They need to add borders on rows or columns, a citation, and align data in cells.  Rows and columns may or may not be titled.
+
+The newsroom also uses Methode/HTML tables for other purposes via templates -- notably timelines and bar charts. These are used less often and are each tied to specific template. For Ellipsis and ANS purposes, they should be considered a separate use case. See [ANS Stat Series](2017-04-20 - Statistic Series.md).
+
+The following proposal deals specifically with the use case of creating a basic data chart.
+
+# Proposal
+
+A new `chart` format that includes row/column settings and a citation, and models cells as text elements (rather than strings.)
+
+## Title
+
+An ANS `text` element.
+
+## Citation
+
+An ANS `text` element, identical to a `citation` on the new `quote` element.  Intended to indicate the source of the data.
+
+## Credits
+
+An ANS credits object, can denormalize authors from author service. Intended to indicate the person who created this specific chart.
+
+## No width/height/sizing options
+
+Explicit sizing options are deliberately not included.
+
+## Explicitly Defined Columns - `columns`
+
+Data columns should be explicitly defined in the `columns` attribute. The following attributes can be set:
+
+* `_id` Automatically-generated.
+* `text_alignment` can be optionally set and applies to the entire column.
+* `title` text element can be optionally supplied here which can function as a header for a column.  **If at least one column has a title than a header row containing the title should be added to the chart. Otherwise it should be omitted.**
+* `borders` an array of strings with the possible values `"left"` and `"right"`
+
+## Explicitly Defined Rows - `rows`
+
+* `_id` Automatically-generated.
+* `title` text element can be optionally supplied here which can function as a header for a row.  **If at least one row has a title then a header column containing the title should be added to the chart. Otherwise it should be omitted.**
+* `borders` an array of strings with the possible values `"top"` and `"bottom"`
+* `cells` an array of cell objects.
+
+## Cells
+
+Currently these objects have just two properties:
+
+* `_id` - Automatically-generated
+* `cell_content` - currently this object can only be an ANS `text` element.
+
+# Example
+
+```
+{
+  "type": "story",
+  "version": "0.5.9",
+  "content_elements": [
+    {
+      "type": "chart",
+
+      "title": {
+        "type": "text",
+        "content": "My Awesome Chart"
+      },
+
+      "citation": {
+        "type": "text",
+        "content": "Source: The Washington Post"
+      },
+
+      "credits": {
+        "by": [{
+          "type": "author",
+          "name": "Angela Wong"
+        }]
+      },
+
+      "columns": [
+        {
+          "_id": "AAAAAAAAAAAAAAAAAAAAAA",
+          "text_alignment": "left",
+          "borders": [ "right" ],
+
+          "title": {
+            "type": "text",
+            "content": "Men"
+          }
+        },
+        {
+          "_id": "AAAAAAAAAAAAAAAAAAAAAB",
+          "text_alignment": "right",
+
+          "title": {
+            "type": "text",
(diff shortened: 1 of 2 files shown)


docs/proposals/2017-04-20 - Charts.md
```diff
@@ -0,0 +1,201 @@
...
+
+* Drop the element.
+
+### Why doesn't this have more formatting options?
+
+ANS Elements should be as narrowly-defined and semantic as possible. The more formatting options that are presented to the user and saved in ANS, the more constrained renderers are in their ability to present the data in an optimal way for their end-users.
+
+### Okay, but why can't I set a column width?
+
+Fixed column widths of any substance make rendering on mobile devices very difficult.  What if a mobile renderer wants to break a table up into multiple parts, or doesn't have as many pixels as the table asks for? Even percentage-based widths can result in columns that are too small to read or cause characters to bleed out. Writers and editors can't be expected to predict every possible format their content will appear in, so better to have them enter the data they want displayed with as much semantic information as possible and let designers  optimize the presentation.
```
StephanieDClark: Methode tables have a column percentage.  Can we add support for that in?

docs/proposals/2017-04-20 - Charts.md
```diff
@@ -0,0 +1,201 @@
...
+* `title` text element can be optionally supplied here which can function as a header for a row.  **If at least one row has a title then a header column containing the title should be added to the chart. Otherwise it should be omitted.**
+* `borders` an array of strings with the possible values `"top"` and `"bottom"`
+* `cells` an array of cell objects.
+
+## Cells
+
+Currently these objects have just two properties:
+
+* `_id` - Automatically-generated
+* `cell_content` - currently this object can only be an ANS `text` element.
```
StephanieDClark: Methode tables often have cell content like this: ```<td align="left" valign="middle" width="25%">
                        <p>label 1</p>
                        <p>
                            <b>Thing 1</b>
                        </p>
                        <p>Description 1</p>
                    </td>```

Note that there are several text elements.  So, the cell_content could actually be its own array of content_elements.

docs/proposals/2017-04-20 - Statistic Series.md
```diff
@@ -0,0 +1,161 @@
...
+
+  "citation": {
+    "type": "text",
+    "content": "Source: The Washington Post"
+  },
+
+  "items":
+  [
+    {
+      "label": {
```
StephanieDClark: This format is the same as the format mentioned above; a table cell can have an array of content elements, e.g., ```<td align="left" valign="top" width="25%">
                                        <p>Column 2</p>
                                        <p class="BigNumber">42</p>
                                        <p>The meaning of life</p>
                                    </td>```

So, why make two separate types for supporting the same type of content?  THis forces the front end to develop multiple jsps or whatever.  When really, a "cell" can have an array of content_elements.

chadmoone: I think the key thing here is that this is *not* a table.  It's a data series, which may be rendered in a table, but other than the direction, there is no implicit formatting.

docs/proposals/2017-04-20 - Charts.md
chadmoone: I can see the argument for having content be an array of `content_elements`.  It feels kind of unfortunate, just because it adds some bloat and will be unused in most cases.

Do we absolutely need `raw_html` here?  I think we'd like to relegate that option to only when and where absolutely necessary, otherwise we will end up with it everywhere.  And it seems like the content of a chart should not be so complex that it can't be represented by text elements.

gengel: My reasoning for not including more than one element here is that I couldn't find a real-world use case for it.  All of the "basic" charts we create (that anyone has shown me, at least) are single-value cells, which is easy to support.  The main exception I'm aware of is the example you gave, but those have a dedicated template in Methode and seem to be pretty clearly a different use case.  (And I proposed an accompanying solution for those)

timothykim: @StephanieDClark Do you have example URLs that contain such tables?

docs/proposals/2017-04-20 - Charts.md
```diff
@@ -0,0 +1,201 @@
...
+          "text_alignment": "right",
+
+          "title": {
+            "type": "text",
+            "content": "Women"
+          }
+
+        }
+      ],
+
```
StephanieDClark: I think this example is meant to represent the table here: https://www.washingtonpost.com/opinions/the-fed-has-a-choice-to-make-with-enormous-stakes/2017/03/12/b7ff95a4-05c5-11e7-ad5b-d22680e18d10_story.html?utm_term=.2829c6fee5d8.  See table below

Note that the "columns" property would really need a colspan to properly render the "header" row.

```<table cellpadding="0" cellspacing="0" width="100%">
    <thead>
        <tr>
            <td align="left" valign="middle" width="33.33%"/>
            <td align="left" valign="middle" width="33.33%">Men</td>
            <td align="left" valign="middle" width="33.33%">Women</td>
        </tr>
    </thead>
    <tr>
        <td align="left" valign="middle" width="33.33%">2000</td>
        <td align="left" valign="middle" width="33.33%">$19.44</td>
        <td align="left" valign="middle" width="33.33%">$15.22</td>
    </tr>
    <tr>
        <td align="left" valign="middle" width="33.33%">2007</td>
        <td align="left" valign="middle" width="33.33%">$19.45</td>
        <td align="left" valign="middle" width="33.33%">$15.90</td>
    </tr>
    <tr>
        <td align="left" valign="middle" width="33.33%">2015</td>
        <td align="left" valign="middle" width="33.33%">$19.18</td>
        <td align="left" valign="middle" width="33.33%">$15.87</td>
    </tr>
    <tr>
        <td align="left" valign="middle" width="33.33%">2016</td>
        <td align="left" valign="middle" width="33.33%">$19.33</td>
        <td align="left" valign="middle" width="33.33%">$16.08</td>
    </tr>
</table>```

docs/proposals/2017-04-20 - Charts.md
StephanieDClark: @timothykim , cc @gengel   Here's an example of a table.  This needs support for a few things that aren't currently in the spec, including:
- valign
- colspan
- width*
- cell content is an array.

*NOTE: I'm open to just passing some of this in additional_properties, eg width.  
I agree also, that I need to build something that is more responsive.  But initially, the information supplied in Methode needs to be passed along.

See attached table, from this article: https://www.washingtonpost.com/business/get-there/no-dont-buy-those-best-stocks-to-own-in-2017/2016/12/10/f7846708-bb39-11e6-91ee-1adddfe36cbe_story.html?utm_term=.5081e4065be5

[Methode-table-1.txt](https://github.com/washingtonpost/ans-schema/files/964825/Methode-table-1.txt)

docs/proposals/2017-04-20 - Charts.md
gengel: You're right, in the HTML you absolutely need to use colspan/rowspan, but you don't need them in the ANS to add them in the HTML.  The ANS object has four rows and two columns. In an HTML `<table>`, this example might be rendered as five rows and three columns -- the extra header rows/columns being inferred from the presence of `title`.  (That's what I was alluding to with "If at least one row has a title than a header column containing the title should be added to the chart. Otherwise it should be omitted." That way the renderer can also decide to change the formatting on the header rows/columns to differentiate them, or use another solution entirely)

docs/proposals/2017-04-20 - Charts.md
gengel: This is a great example, thank you!

gengel:
Gregor Team to write a new proposal based off of @timothykim 's rough draft

PR #132, opened 2017-04-24
Fixes bug in upverter where empty string status code gets converted t…

gengel:


gengel:
@jpblair I am merging now but can you please verify this

PR #134, opened 2017-04-27
Add text formatting proposal

vgritsenko:
~This PR is strictly for shaping the initial proposal within the text attributes working group, and a new one will be set up to handle actual implementation of changes as well as acceptance  by the ANS committee as a whole.~

_Update 8/16/2017 (via @chadmoone):_ This PR has developed and will be looked at as the formal proposal for this change.  Once this proposal is formally accepted and merged, implementation will begin.

### DIFF
--- a/docs/proposals/2017-04-27 Text Formatting.md
+++ b/docs/proposals/2017-04-27 Text Formatting.md
@@ -128,6 +128,18 @@ Additionally, `link` attribute must have `url` property pointing to the linked c
         ]
 ```
 
+This initial list of attributes can be extended at a later versions of the specification with additional attributes, or with additional properties on existing attributes, if such need arises. For example, `strong` attribute could be extended to include `weight` property:
+
+```
+        "attributes": [
+          {
+            "_id": "AAAAAAAAAAAAAAAAAAAAAAAAAC",
+            "type": "strong",
+            "weight": 500
+          }
+        ]
+```
+
 ### `content` is an HTML representation of the canonical `text` property
 
 Each attributed text object can have an optional `content` element which contains HTML representation of that `text`. It is expected that, on input:
@@ -168,9 +180,12 @@ The exception is the `code/content` property, which represents plain text comput
 
 Existing properties which allow usage of the markup in text content should adopt `attributed_text` type instead of `string`. These include:
 
+* author/bio
 * image/caption
+* tag/description
 * video/transcript
 * description (trait_description.json / dictionary.json)
+* editor_note (via trait_editor_note.json)
 * headlines (trait_headlines.json / dictionary.json)
 * subheadlines (trait_subheadlines.json / dictionary.json)
 
@@ -199,7 +214,6 @@ All other `string` type properties are assumed to have only plain text content. 
 * code/language
 * address/street_address (via trait_address.json)
 * copyright (via trait_copyright.json)
-* editor_note (via trait_editor_note.json)
 * location (via trait_location.json)
 * label/basic/text (via trait_label.json)
 * label/*/text (via trait_label.json)
@@ -208,14 +222,12 @@ All other `string` type properties are assumed to have only plain text content. 
 * source/name (via trait_source.json)
 * author/name
 * author/org
-* author/bio
 * auxiliary/name (taxonomy)
 * keyword/keyword (taxonomy)
 * named_entity/name (taxonomy)
 * site/name
 * site/description
 * tag/text
-* tag/description
 * topic/name
 * video/rating
 
@@ -224,3 +236,4 @@ All other `string` type properties are assumed to have only plain text content. 
 
 This proposal leaves room for defining other types of attributes, such a `link`, which could be used to denote relationship between elements and other content.
 
+Determine if site/description and image/subtitle could contain attributed text.
--- a/docs/proposals/2017-04-27 Text Formatting.md
+++ b/docs/proposals/2017-04-27 Text Formatting.md
@@ -64,7 +64,7 @@ Example:
   }
 ```
 
-### `selection` describe formatting for specific segments of a text string
+### `selection`s describe formatting for specific segments of a text string
 
 Each `selection` object defines a part of a text string and what formatting attributes should be applied to it. An idea borrowed from Apple and Twitter.
 
@@ -115,6 +115,26 @@ Example:
         ]
 ```
 
+### `content` is an HTML representation of the canonical `text` property
+
+Each attributed text object can have an optional `content` element which contains HTML representation of that `text`. It is expected that, on input:
+
+* `text`, and `attributes`, and `selections` properties are present, or recreated from the `content` property during import;
+* `content` property is omitted, or contains only allowed subset of the HTML;
+
+On output:
+
+* `text`, and `attributes`, and `selections` properties are present;
+* `content`  property can be omitted for non HTML clients;
+
+This means that ARC producers, when HTML content is passed in, should validate and parse that HTML content into attributed text content. On the other hand, HTML centric consumers can ignore attributed text markup and use raw HTML representation directly.
+
+Supported HTML subset will consist of tags and styles which can be directly mapped to the attribute types:
+
+* for strong, `<b>` and `<strong>`,
+* for emphasis, `<i>` and `<em>`,
+* etc.
+
 
 ## Part II. Structured Text Applied to Story Elements
 
@@ -148,14 +168,16 @@ Existing properties which use `text` where additional properties defined in `con
 * table/rows/items/items
 
 
-## Part IV. Document Usage of HTML
+## Part IV. Document Usage of Raw HTML
 
-### Define properties allowing HTML content
+### Define properties allowing raw HTML content
 
 All `string` type properties which can have HTML content should be explicitly documented. These include:
 
 * raw_html/content
 * video/embed_html
+
+### Define properties prohibiting HTML content
 
 All other `string` type properties are assumed to have only plain text content. Examples include:
 
--- a/docs/proposals/2017-04-27 Text Formatting.md
+++ b/docs/proposals/2017-04-27 Text Formatting.md
@@ -42,11 +42,11 @@ It should also:
 
 # Proposal
 
-## Part I. Structured Text Element.
+## Part I. Structured Text Object.
 
-### `text` story element should describe structured text content
+### `attributed_text` object describes structured text content
 
-Text element will have following structure:
+Attributed text object will have the following structure. Note that proposed structure is compatible with preexisting `text` story element (as of 0.5.8).
 
 * a required `text` property with plain text content of type `string`,
 * an optional `attributes` property with array of `attribute` objects describing formatting to be applied to the text,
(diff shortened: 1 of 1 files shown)


docs/proposals/2017-04-27 Text Formatting.md
```diff
@@ -0,0 +1,176 @@
...
+
+
+## Part II. Structured Text Applied to Other Elements
+
+### All formatted content should be represented as a structured text content
+
+Existing `blockquote`, `code`, `correction`, `header`, `interstitial_link` adopt the same structure as a `text` element. Following properties are replaced with text/attributes/selections/text properties:
+
+* blockquote/content
+* code/content
```
gengel: I think `code` should *not* be included in this list

docs/proposals/2017-04-27 Text Formatting.md
```diff
@@ -0,0 +1,176 @@
...
+
+All `string` type properties which can have HTML content should be explicitly documented. These include:
+
+* raw_html/content
+* video/embed_html
+
+All other `string` type properties are assumed to have only plain text content. Examples include:
+
+* image/subtitle
+* image/caption
```
gengel: I believe that headlines, subheadlines, and image captions all occasionally make use of em/italics.  Should they be included in the formatted text list above?

docs/proposals/2017-04-27 Text Formatting.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Part IV. Future Work
+
+This proposal leaves room for defining other types of attributes, such a `link`, which could be used to denote relationship between elements and other content.
+
+
+# OPEN QUESTIONS #
+
+* Should text (trait_label.json) support formatting/HTML?
+* Should main (train_disctionary.json) support formatting/HTML?
```
gengel: Arguably these should be dropped and the fields that use them should each define their own schema -- they are more utility groupings for schema than actual semantic traits

But my hot take is:
  - labels: no formatting
  - headlines, subheadlines, description: formatting

docs/proposals/2017-04-27 Text Formatting.md
```diff
@@ -0,0 +1,176 @@
...
+        ]
+```
+
+
+## Part II. Structured Text Applied to Other Elements
+
+### All formatted content should be represented as a structured text content
+
+Existing `blockquote`, `code`, `correction`, `header`, `interstitial_link` adopt the same structure as a `text` element. Following properties are replaced with text/attributes/selections/text properties:
+
```
gengel: Is this preferable to simply giving these elements (or new replacement elements) their own `content_elements` property (perhaps with a max size of 1, allowed type of only `text`)?

vgritsenko: `content_elements` imply that contents must be a `content_element` which brings everything defined in it:
* channels
* alignment
* ...

As a result, there could be a `code` element with one value of `alignment` and then nested `text` with another value of `alignment` which is probably - as I understand it - is not an effect we are going here for.

The right option - an alternative option - would be to define another class - let's call it `attributed_text` for the lack of the better name - and using it here and in every other place. This alternative does come with a cost of adding yet another layer to the ANS onion.

docs/proposals/2017-04-27 Text Formatting.md
vgritsenko: The platform on which you are reading this proposal and left this comment automatically highlights the code syntax. Shouldn't we support this option too?

docs/proposals/2017-04-27 Text Formatting.md
vgritsenko: Ok. This would require an `attributed_text` trait to be defined.

docs/proposals/2017-04-27 Text Formatting.md
gengel: Ok, I'm convinced

docs/proposals/2017-04-27 Text Formatting.md
gengel: The key word here being "automatically." This seems like a decision better left to the renderer -- I don't have to manually mark up my own coloring in github when I post a code sample

vgritsenko: Ok, I see your line of reasoning.

gengel:
Meeting 7-13: Replace attributes array with named fields in a dictionary,
each attribute needs schema, name

gengel:
Fields to move to attributed text:

site/description ?
tag/description
author/bio
editor_node
image/subtitle ?

chadmoone:
After discussing with @vgritsenko some more, I think he was correct to set the attributes up as an array rather than named fields, so we are leaving that part of the proposal as-is.

calexicoz:
This looks good to me, though I have a couple random questions/comments.

My particular use case that I will want to address in ANS at some point is "annotations". I have this app which automatically scans the content of articles for dedicated keywords, and which can then link those keywords to additional content (this can result in an actual link being created, or in an "in-page" popup providing some context around the concept/person/event/place/...).

I don't think that my use case fits in the general pattern of "formatting" and therefore I wouldn't want to use the current proposed elements (in particular because those contextual elements are added after the fact, and so it would be easier to deal with them in a separate section which can be updated separately from the content).

That being said, the elements I would require to describe annotations would pretty much be identical to the one used for formatting: text selections with attributes.

Anyway, this is really more a sideway comment, and I don't think it affects this proposal directly. I don't know much about ANS yet,  it's just that the use cases for formatting and annotations are fairly similar, and it made sense to maybe at least mention it. I'll probably have to dig deeper into ANS at some point to see where/how the annotation issue can be addressed.

Hope this is at least somewhat helpful.

chadmoone:
Thanks for the feedback @calexicoz!

Actually, I almost modified the title of this to use the term 'Annotated' or 'Attributed' text rather than 'Formatting', because we absolutely intend that it could be extended to link various metadata and other attributes to selections of text.  Examples could be an editor's note on a selection of text, a stock ticker info added to a company name, etc.  It would be stored in this way inside of ANS, and then the client can optionally use that data to render differently or just disregard if unsupported.

Though we haven't included the actual spec here to clarify (though maybe we should amend the proposal), I think the intention is that arbitrary attribute objects could be added into the `attributes` array for a selection, so you would be able to store most anything you want there. However, it's probably ideal to add an official spec for any new attributes whenever possible, to allow other consumers to make use of them and to prevent a shadow spec. New attributes should be easy to add, because they would generally be additive and non-breaking.

There is also the case that has come up where some of this info would be generated on the client side entirely, and not actually stored inside of the content API. In this case, the implementation is kind of outside of the scope of the ANS spec, but certainly it could follow this pattern.

If you think it would be helpful, we could include an example in this proposal. Or we could add language to clarify this. Let me know if you have any thoughts/suggestions on this, or send a PR on this branch if you want to add something.

calexicoz:
I think the main concern for annotation vs formatting is "when does it happen". Let me clarify.

When you deal with formatting, the formatting is known at the time of content generation. Formatting is generated by the editor (say Ellipsis). So it makes sense to have the formatting info alongside the content.

For annotations, it is possible, and it's the case for me, that those are created as a separate step, part of an automatic process. In the case of the Content API, I don't think we have solved the problem of "multiple parties can update the content itself". It seems that the general approach regarding Content API and ANS has been: multiple parties collaborate with each other by each dealing with their own section. I.e. there is some form of ownership of different part of the ANS object, and only one party updates a particular section at any point in time.

So, the problem really isn't around the proposal itself. I would 100% agree that the current proposal, from a "JSON schema" standpoint, works perfectly for annotations. I would also agree that we should probably have a more formal definition of the range of attributes allowed under selections.

The problem is really more: if we allow annotations as part of this, how will multiple systems interact with the ANS object in a collaborative manner, and not overwrite each other.

This may be more a problem of Content API, and not the ANS spec proper, but I think it's a fairly big (and I think unsolved at this point) problem.

This is sort of why I was hinting at maybe have a similar (identical even) section, but just separate from the content itself, to facilitate the interactions between systems.

robpc:
Just dropping a note here that is a discussion ever happens on @calexicoz 's tangental question/topic/issue, it's been something I've been thinking about a lot for Rainbow and would love to be included in that discussion.

docs/proposals/2017-04-27 Text Formatting.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+### `content` is an HTML representation of the canonical `text` property
+
+Each attributed text object can have an optional `content` element which contains HTML representation of that `text`. It is expected that, on input:
+
+* `text`, and `attributes`, and `selections` properties are present, or recreated from the `content` property during import;
+* `content` property is omitted, or contains only allowed subset of the HTML;
+
+On output:
```
jperezlatimes: Does this mean that both `text` and `content` are used to display text in a published story depending on the platform the story is being read on (Apple News, AMP, a web browser)?

chadmoone:
Both probably wouldn’t be used at the same time by the same client, but either one could be used by a given client. Probably what you meant, but just clarifying.

The canonical representation would live in the text object, but a HTML version would also be presented for legacy use or other cases where it’s simpler to use an HTML representation. This way, we also have a single, canonical fallback representation, and every HTML client does not need to implement their own translation.


-- Chad

docs/proposals/2017-04-27 Text Formatting.md
```diff
@@ -0,0 +1,239 @@
...
+    "type": "text",
+    "text": "The text attribute should contain plain text content only.",
+    "selections": [ ... ]
+    "content": "The <b>text</b> attribute should contain <i>plain text</i> content only."
+  }
+```
+
+### `selection`s describe formatting for specific segments of a text string
+
+Each `selection` object defines a part of a text string and what formatting attributes should be applied to it. An idea borrowed from Apple and Twitter.
```
jperezlatimes: If I have text that I need to show up in an editor app but I don't want displayed in a published story, such as a note between a copy editor and a reporter, would there be a way to denote that with this schema?

docs/proposals/2017-04-27 Text Formatting.md
```diff
@@ -0,0 +1,239 @@
...
+
+* a required `start` and `length` integer properties define sub-string range,
+* an optional `attributes` property with array of `attribute` objects.
```
jperezlatimes: If I have text that I need to show up in an editor app but I don't want displayed in a published story, such as a note between a copy editor and a reporter, would there be a way to denote that with this schema?

gengel: My bias is that content that isn't directly or indirectly reader-facing should not be included in ANS. But you could certainly save+update a "notes" document that keys off of a (story id + revision id + content element id) and is loaded in an editor alongside a story revision

calexicoz: That's my feeling as well. This kind of notes would be very similar to "annotations", and this proposal is really about formatting. Those should live in separate places. Though, I would agree that the technical requirements would be very similar. There is still the concept of a text selection. Maybe we need to make a separate proposal to address those use cases.

chadmoone: I agree with the distinction between reader-facing vs private/internal, but I do think that this spec should be updated to support annotations which _are_ reader-facing but not specifically 'formatting' (as in bold, underline, italic, etc).

For example, we've discussed the new 'notes' story tools feature, which allows a popup document to be displayed when they reader hovers or taps on a certain selection of text. This is sort of similar to the 'Genius' notes system that The Fix was using last year.  My preference is to have something like this included in the `attributes` along with other formatting and presentation attributes.  

Another good reason to separate internal/private annotations is that we probably want to avoid those being sent to user-facing clients in the JSON document, even if the client doesn't render them. Exposing something like editor notes even in the JSON API could potentially lead to an embarrassing situation.

So for me, the distinction is really more of an access/visibility concern, rather than the type of attribute/annotation.

chadmoone: Though, I should add that this also gets to the point @calexicoz brought up above, in the case that these annotations are being added by a separate tool from the original editor.  This, though, I think is honestly more of a problem that needs to be solved on the content API side as opposed to the schema.

gengel:
Add notes about special characters / internationalization

gengel:
Draftjs - does it have a fixed schema?

Can you limit supported features?
If not, how do we handle unsupported data types?

calexicoz:
Hey there,

I stumbled upon the Mobile Doc format a couple times, and it as an interesting approach to this problem. I figured it would be good to at least take a look at it, to see how other solve this problem. So this is more a reference for later: https://github.com/bustle/mobiledoc-kit/blob/master/MOBILEDOC.md

michaelguild13:
Whats the status of this proposal? its 2019 :)

chadmoone:
Right... My understanding is it's basically 'approved but continuously delayed'.

gengel:
@chadmoone @michaelguild13 We're going to work it into the new platform. I'll try to get you details soon

PR #135, opened 2017-04-28
Reworks trait_publishing to make elasticsearch 1.5 happy

gengel:
As I started working on the equivalent ES index updates for trait_publishing, I realized that in 1.5 (which is the current ES version deployed for most clients) the schema we had agreed upon earlier was going to have some field collisions in certain queries that would be Not Good.  This PR largely ditches the attempts to make trait_publishing fields match the story_operation (Content Ops) type and the edition type used internally in Story API and goes for prefixed fields names instead, also adding additional_properties to both objects. 

The goal here was make sure that the queries WebSked will need will work, and also that existing queries that rely on the "published: true" query that works in 1.5 and nowhere else will not suddenly start returning false positives. 

This is somewhat bad form, as I'm effectively trying to squeeze in one non-trivial 0.5.8 change after it's gone out, but no one's using the changed fields yet so it should be ok.

See also: https://github.com/WPMedia/arc-content-api-tasks/pull/103/files

### DIFF
--- a/src/main/resources/schema/ans/0.5.8/traits/trait_publishing.json
+++ b/src/main/resources/schema/ans/0.5.8/traits/trait_publishing.json
@@ -139,22 +139,23 @@
           "format": "date-time"
         },
 
-        "edition_publish_status": {
-          "title": "Publish Status",
+        "edition_published": {
+          "title": "Publish Status (Edition)",
           "description": "If false, this edition has been deleted/unpublished.",
           "type": "boolean"
         },
 
         "edition_revision_id": {
-          "title": "Revision ID",
-          "description": "The id of the revision that this edition was created from. Omitted if unpublished.",
+          "title": "Revision ID (Edition)",
+          "description": "The id of the revision that the edition was created from. Omitted if unpublished.",
           "type": "string"
         },
+
         "additional_properties": {
           "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.5.8/traits/trait_additional_properties.json"
         }
       },
-      "required": [ "edition_publish_status", "edition_date", "edition_name" ]
+      "required": [ "edition_published", "edition_date", "edition_name" ]
     }
   }
 }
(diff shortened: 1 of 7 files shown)


src/main/resources/schema/ans/0.5.8/traits/trait_publishing.json
```diff
@@ -96,40 +118,43 @@
...
-        "publish_date": {
+        "edition_publish_date": {
           "title": "Publish Date (Edition)",
           "description": "The machine-editable date that should be shown to readers as the 'publish date' for this content item. When viewing the story at this edition name directly, this will override whatever value is set for Publish Date on the story directly. Every time an edition is updated (i.e. a story is republished) this date will also be updated unless otherwise specified.",
           "type": "string",
           "format": "date-time"
         },
 
-        "published": {
+        "edition_publish_status": {
```
jpblair: Since this is now namespaced, should it go back to `edition_published`?  Now that I look, that probably makes more sense for a boolean value - it's not that clear what `edition_publish_status=true` means.

src/main/resources/schema/ans/0.5.8/traits/trait_publishing.json
```diff
@@ -40,50 +40,72 @@
...
+                "type": "string"
+              },
+              "additional_properties": {
+                "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.5.8/traits/trait_additional_properties.json"
               }
-            ]
+            }
           }
         },
         "unpublish_edition": {
```
jpblair: Should this have `operation_revision_id` like `publish_edition` does?

tests/fixtures/schema/0.5.8/story-fixture-tiny-house.json
```diff
@@ -141,41 +141,35 @@
...
+        "edition_publish_status": true
       }
     },
     "scheduled_operations": {
       "publish_edition": [{
-        "type": "story_operation",
-        "story_id": "f8f706f0-0acc-11e5-9e39-0db921c47b93",
         "operation": "publish_edition",
-        "date": "2017-04-20T15:00:00Z",
-        "publish_date": "2017-04-15T15:00:00Z",
```
jpblair: It looks like a few fields have gone away here (`publish_date`, `display_date`) - do we need them for anything?  I think you said WebSked would only ever be concerned with `operation_date`, but just want to be sure.

PR #137, opened 2017-05-01
Specify URI format for all URL fields

chadmoone:
This adds strict`uri` format validation to all URL fields, as per [RFC3986](http://tools.ietf.org/html/rfc3986).  Both absolute and relative (reference) URI will pass validation.

All tests pass.  No new tests have been added, since they would essentially be testing JSON Schema validation.

A few fixtures were updated to pass tests, though they were all for the `reference.provider` field, which will be deprecated/removed soon.

### DIFF
--- a/src/main/resources/schema/ans/0.5.9/traits/trait_label.json
+++ b/src/main/resources/schema/ans/0.5.9/traits/trait_label.json
@@ -17,7 +17,8 @@
         },
         "url": {
           "type": "string",
-          "description": "An optional destination url of this label."
+          "description": "An optional destination url of this label.",
+          "format": "uri"
         },
         "display": {
           "type": "boolean",
@@ -42,7 +43,8 @@
         },
         "url": {
           "type": "string",
-          "description": "An optional destination url of this label."
+          "description": "An optional destination url of this label.",
+          "format": "uri"
         },
         "display": {
           "type": "boolean",
--- a/src/main/resources/schema/ans/0.5.9/traits/trait_social.json
+++ b/src/main/resources/schema/ans/0.5.9/traits/trait_social.json
@@ -11,7 +11,8 @@
         "type": "string"
       },
       "url": {
-        "type": "string"
+        "type": "string",
+        "format": "uri"
       }
     },
     "additionalProperties": {}
--- a/src/main/resources/schema/ans/0.5.9/utils/reference.json
+++ b/src/main/resources/schema/ans/0.5.9/utils/reference.json
@@ -44,7 +44,8 @@
         },
         "provider": {
           "description": "A URL that can resolve the id into an ANS element",
-          "type": "string"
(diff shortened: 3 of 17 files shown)


src/main/resources/schema/ans/0.5.9/utils/reference.json
```diff
@@ -44,7 +44,8 @@
         },
         "provider": {
           "description": "A URL that can resolve the id into an ANS element",
-          "type": "string"
+          "type": "string",
+          "format": "uri"
```
gengel: I know this seems wrong, but I suspect we may have to leave `provider` alone, if only because a *lot* of code out there stubs in the empty string for a provider (owing to its being a required field, also wrong)

Possibly we could replace this with:


```
 "type": "string",
 "oneOf": [
   { 
     "enum": [ "" ]
    },
    {
      "format": "uri"
    }
  ]
```

chadmoone: Ok.  This serves as a pretty good argument for getting fields like this validated from the start.

This alternative, while not the prettiest thing in the world, passes all tests and would allow the old content to be grandfathered.

There was also talk about deprecating/removing this field altogether—should that just happen for 0.5.9, and then we can leave this unvalidated?

src/main/resources/schema/ans/0.5.9/traits/trait_label.json
```diff
@@ -17,7 +17,8 @@
         },
         "url": {
           "type": "string",
-          "description": "An optional destination url of this label."
+          "description": "An optional destination url of this label.",
+          "format": "uri"
```
vgritsenko: Nitpicking. Shouldn't `format` come before `description`?

src/main/resources/schema/ans/0.5.9/traits/trait_social.json
```diff
@@ -11,7 +11,8 @@
         "type": "string"
       },
       "url": {
-        "type": "string"
+        "type": "string",
+        "format": "uri"
```
gengel: "url" is unfortunately something of a construct here -- I think we need to drop it for now for backwards compatibility

src/main/resources/schema/ans/0.5.9/utils/reference.json
gengel: I agree it's better to be strict from the start.

We can't drop the field altogether, as it's needed for oembed objects.  The best we can do, I think, is make it non-required for all other reference types and force it into the restrictions above

chadmoone:
I think the principal of this is still valid, but it's way out of date. @timothykim @gengel is this worth recreating? Is your read that with a few caveats like mentioned above, this is something that we can bring in?

Sorry for letting this fall so far out of date. Just has not been front of mind.

PR #139, opened 2017-05-08
Proposal for tweet element

gengel:
A proposal for a discrete `tweet` element distinct from `oembed_response` and `raw_html`.

Desired for Arc Voice Transcription project


timothykim:
This might be a dumb question but is a single tweet denormalized in all the stories that reference the tweet?

gengel:
Yes, if inserted as a reference

PR #143, opened 2017-05-25
2017-05-24 - Update Revision to include editor

epascarello:


gengel:
Is `.revision.editor_id` intended to have the same possible values as `.source.system`? If so, why not `revision.system`?

epascarello:
I do not care what the property is named, I just want something that I can reference when people say XYZ version has an issue and I want to see if it was Ellipsis or something else. I honestly do not like editor_id since it sounds like who (user) edited it.

gengel:
Do you still want this change?

PR #144, opened 2017-06-12
Advanced Table Proposal

jmjakes:
This PR contains a proposal for a new table type, `advanced_table`.

* needed some way to support tables from methode (other option is raw html)
* wanted something that can be converted from html, and can be clearly interpreted by clients
  * this means turning it back into html, or whatever markup the client is using
* it intentionally encapsulates basic functionality that exists on an html table, but is intended to be implementation agnostic
  * doesn't contain any implementation details past implying the basic layout, so it is up to clients to determine how to render it

### DIFF
--- a/docs/proposals/2017-5-26 - Advanced Table.md
+++ b/docs/proposals/2017-5-26 - Advanced Table.md
@@ -1,0 +1,444 @@
+# Problem
+
+Current ANS tables are not sufficient to support tables from methode.
+
+
+# Proposal
+
+Add a new table spec that mirrors the functionality of an HTML table, called `advanced_table`.
+
+Key traits of the table:
+
+* 4 data types - `advanced_table`, `row`, `data_cell`, `header_cell`
+  * a `table` is a list of rows
+  * a `row` is a list of cells
+  * a `cell` is a list of content elements, restricted to text and raw_html
+    * there are two implementations of a cell, `data_cell` and `header_cell`, similar to `<th>` and `<td>` tags
+* order/layout assumptions - elements will have an assumed layout that clients may follow
+  * for example, it is assumed that rows are ordered from top to bottom, and cells are ordered left to right
+* no other rendering details will be included in the spec
+  * adapters can use additional properties to specify rendering details, if desired
+
+
+## Advanced Table
+
+`advanced_table` is the parent object. It will be a content element, and contain a list of rows.
+
+### Rows
+
+List of rows that belong to the table. The order is assumed to be top to bottom.
+
+### Title
+
+ANS field. Name of the table.
+
+### Citation
+
+ANS text element. The source of the table, for example "The Washington Post".
+
+### Credits
+
+ANS credits object, for identifying the author(s) of the table.
+
+### No height/width or other display options
+
+These are intentionally left out.
+
+If adapters need to include any style or rendering details, they can use additional properties on any component of a table - table, row, cell, or content element.
+
+
+## Row
+
+A `row` contains a list of cells. It will only be usable within a table.
+
+### Cells
+
+List of cells that belong to this row. The order is assumed to be left to right.
+
+
+## Data Cell / header Cell
+
+There are two types of cells, `data_cell` and `header_cell`. They will only be usable within a row.
+
+Both `header_cell` and `data_cell` will contain a list of content elements. They have the same spec, except for the type.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/2017-5-26 - Advanced Table.md
```diff
@@ -0,0 +1,446 @@
...
+### Cell types
+
+The spec could potentially have one type for cells, `cell`, and use subtype to distinguish between `data` and `header` cells. This would make the schema smaller, but would also require clients check two fields on each cell. I don't have strong feelings on this either way.
+
+
+## Concerns
+
+### What happens if rows have different numbers of cells?
+
+Based on the order of elements and the assumed layout (left to right, or top to bottom), it is clear where each element should go. If the user wants to leave empty space, they would include empty elements. Otherwise, it will be assuemd that any mismatch in length will lead to space at the end of the element.
```
gengel: To make sure I understand this:

Row 1 has 2 cells, 
Row 2 has 5 cells

When rendering, the cell 2 in row 1 should have the same width as the total width of cells 2-5 in row 2

Is that correct?

docs/proposals/2017-5-26 - Advanced Table.md
```diff
@@ -0,0 +1,446 @@
...
+### Subtype
+
+ANS subtype. Can be used to distinguish between different types of tables, so that clients can react accordingly.
+
+### No height/width or other display options
+
+These are intentionally left out.
+
+If adapters need to include any style or rendering details, they can use additional properties on any component of a table - table, row, cell, or content element.
+
```
gengel: The risk of leaving display options in additional_properties is that they may be ignored or deleted by apps that access the data later and haven't specifically been coded to deal with the Methode options.  

For Ellipsis, the scenario would be:

1) User creates a table in a story in Methode
2) Story is ingested into Content API with lots of custom display options in `additional_properties`
3) Later, a different user opens the story in Ellipsis,  moves some table cells around, saves the document
4) Ellipsis saves the "new" table that the user edited in place of the old table, deleting all Methode formatting

There are also implications for native apps. It seems to me that this requires one of the following to be true:

1) Native apps coordinates with Gregor and implements rendering solutions for all possible additional_properties that might be added to a table
OR
2) Display options set in additional_properties are not present in native app renderings 

Perhaps this is fine, but it's worth flagging as a consequence of this approach.

docs/proposals/2017-5-26 - Advanced Table.md
```diff
@@ -0,0 +1,446 @@
...
+Raw html makes it possible to include more advanced markup, such as span tags, on elements within a cell. Raw html would be used sparingly, only when text elements are not sufficient.
+
+While this currently isn't a problem, because text elements do not have restrictions, my understanding is that this will change in the future. I believe allowing raw_html in a cell will future proof this functionality, and avoid unnecessary migrations.
+
+An example from methode:  
+```
+<p><span>METHODOLOGY</span> \"MOSTLY, VALUE OVER MOMENTUM STOCKS\"</p>
+```
+
```
gengel: If a table might contains raw html that cannot be represented as `text`, why not just render the whole table as a `raw_html` element? A table with 1 or more raw_html elements requires downstream consumers to be able to parse and render HTML to render the table in its entirety.

docs/proposals/2017-5-26 - Advanced Table.md
jmjakes: My intention was that, when rendered, there would be three empty cells at the end of the row 1. Ultimately, this could work either way.

docs/proposals/2017-5-26 - Advanced Table.md
jmjakes: This is probably the strongest reason for us to go the route of raw_html. I don't think we will be able to get around this problem without including a significant amount of rendering specification in the spec.

jmjakes:
@gengel appreciate the feedback, you bring up a lot of good points that I don't really have any good solutions.

While I think it was probably worth my time to write this up, you're feedback is confirming what we are all probably thinking - we should probably just use raw_html. Methode tables are going to more like legacy content moving forward, and maybe it just doesn't make sense to try to squeeze them into ANS.

jmjakes:
Gregor team has agreed to go with raw_html for tables.

PR #147, opened 2017-08-03
Proposal for source event metadata on outgoing Content API updates

gengel:
Schema change for
https://arcpublishing.atlassian.net/browse/CA-108

Related to:
https://arcpublishing.atlassian.net/browse/CA-109
https://arcpublishing.atlassian.net/wiki/spaces/CAID/pages/49595454/WebSked+Duplicate+Tasks


jpblair:
Could this be confused with the `source` field in ANS?  Looks good otherwise.

gengel:
Good point -- would `source_operation` be better? `upstream`? `trigger`?

gengel:
Updated to replace `source` with `trigger`

PR #148, opened 2017-09-25
ANS Author Proposal

gengel:
An attempt to formalize the new transparency fields (and other author service data) in ANS in a consistent way.

### DIFF
--- a/docs/proposals/2017-09-25_Author.md
+++ b/docs/proposals/2017-09-25_Author.md
@@ -1,0 +1,95 @@
+## Problem
+
+Author objects currently are a very small set of the fields available to configure via the Arc Author Service/API. Most of the data from Author Service is available on via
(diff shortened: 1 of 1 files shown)


gengel:
The existing endpoints will continue to serve the existing fields and all additional_properties currently populated will continue to be populated so nothing will break.  But if you switch to the new API endpoints, you'll get the ANS data structure (which has always used `name`), and the new fields will begin to be denormalized into document bodies

docs/proposals/2017-09-25_Author.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Proposal
+
+Add fields to the current ANS author object
+
+```
+{
+  // Current fields
+  "type": "author",
+  "name": "Gregory Engel",   // Equivalent to byline
```
gengel: Nope

PR #149, opened 2017-10-16
Adds fields for priority queue support

gengel:
For the forthcoming priority queue feature in Content API, a way is needed to distinguish the priority of source events for priority ordering. The optional `priority` field is being added to all operation types to distinguish ingestion events from normal usage events. To ensure that this feature is used for its intended purpose, `app_name` will also be required for events requesting to set their own priority level. This field can be mapped to whitelists or blacklists by operation consumers.


gengel:
This PR is relevant for anyone who consumes events from Content API or the primary source services.

jpblair:
Separating the Kafka topics was the main concern from WebSked's perspective.  Metadata in the message itself didn't really help our latency problem, since we were still using processing power to consume the messages, even if we ended up throwing them out.  So under the current architecture, I'm not sure if this has much of an impact for us.

My only comment would be that `app_name` doesn't feel that descriptive.  Perhaps something like `op_initiator` or `op_requester`?

gengel:
`app_name` was chosen to match HTTP header we are requiring incoming requests to specify themselves ("Arc-App-Name") which is matched against a whitelist of applications that are allowed to use the "fast lane" (standard priority)

jpblair:
And that whitelist is configured within the Content API?  The system as it is with the lite topics works for us, so as long as that stays in place, this looks fine to me.

PR #151, opened 2017-11-17
Adds `rating` and `endorsement` element proposal

gengel:


docs/proposals/2017-11-17_-_Ratings.md
```diff
@@ -0,0 +1,163 @@
...
+  "max": 5.0,
+  "units": "stars"
+}
+```
+
+## New 'Endorsement' element
+
+A new content element `endorsement` will be added to indicate text-only appraisals
+
+* `endorsement`: a string indicating the value (e.g. "Buy" or "Don't Buy") (required)
```
timothykim: Instead of having a unsortable endorsement field, why don't we add a displayName so that the translation of score text to Text value (like from the thumbs up and thumbs down example) doesn't have to happen at the rendering layer.

gengel: I'm fine adding displayName, but I do think it's worth demarcating the difference between quantifiable / unquantifiable ratings

timothykim: I can't think of a rating system where the rating is unquantifiable. Can you provide an example?

gengel: WashPost Pinnochio system

awong7: It's up to the org to decide their ratings. So for books, you could have ratings of must read, don't read, maybe read. Or for movies go see or don't see.

docs/proposals/2017-11-17_-_Ratings.md
```diff
@@ -0,0 +1,117 @@
...
+    {
+        "type": "text",
+        "content": "Disney's <em>Beauty and the Beast</em> is pretty good, all things considered."
+    },
+    {
+        "type": "rating",
+        "rating": 4.5,
+        "min": 0,
+        "max": 5,
+        "units": "stars"
```
chadmoone: Do we need to include this here, and if so, should it be displayable text?

gengel: "Units" is mainly there to disambiguate different types of ratings... 

A single organization might rate movies n out of five stars, but rate lies as p out of four pinnochios.  So you could right a generic text renderer that just output ({rating} out of {max} {units}), or you might check specifically for "stars" and then go load a bunch of star pngs.  

As for whether or not to make it explicitly about reviews -- I'm not sure.  I think it's reasonable to use them for rating things that aren't *exactly* reviews, but I agree that they could be confused with other uses of the word "rating."  (And ironically, an age-appropriateness rating is better captured by "endorsement")

For lack of any better terms, I'm inclined to leave them as is, but I'd love to hear any suggestions

PR #152, opened 2017-11-17
Adds `infobox` proposal

gengel:


awong7:
Got feedback from GAM that querying doesn't sound feasible with this option and there is still interest in potentially querying info in the infobox and may come back up again in 2018. Definitely could see this being asked as well from other partners. There are also outstanding questions of whether this is just an interest from partners or something that a partner is committed to building out and this being an absolute need.

gengel:
Seems like there is a general lack of enthusiasm for this one so I'll close

PR #153, opened 2017-12-01
Updated source.source_type with lowercase "staff" and "wires" examples

charlex:
Filing this, as WebSked is expecting case sensitive "staff" and "wires" values for its filtering. "Staff" and "Wires" will not work.


gengel:
Note that this is just examples, it will not change the validation

PR #154, opened 2017-12-08
Adds rating element to ANS schema CA-317

jodieburnett:
# What is implemented?

Adds a new 'rating' element to the ANS schema.

# How was it tested?

Unit tests were added for the new schema, and a test that validates an entire story was modified to include the new element. The modified test still passes with the new element added. 

# What are the side effects or dependencies?

Deploying changes to ANS can be tricky, will have to come up with plan for deployment.
(Any obvious side effects that I am missing?)

# What are the failure cases?

None that I know of. 

# What metrics were added?

n/a

# Where is the documentation?

Repo itself is documentation for ANS schema.

# Did you update RELEASE NOTES?

Yes

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,8 @@
 
 ### TBD ###
 
+* Adds `rating` element
+
 ### 1.3.3 2017/05/16 ##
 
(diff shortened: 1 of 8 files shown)


gengel:
LGTM

PR #157, opened 2017-12-18
creating proposal for ans change for video subtitles

williamslm:


docs/proposals/2017-12-18 - Video-Subtitle.md
```diff
@@ -0,0 +1,69 @@
...
+```
+"subtitles": {
+      "confidence": 1.0,
+      "urls": [
+        {
+          "format": "WEB_VTT",
+          "url": "https://closedcaptions.posttv.com/2017/12/13/5b593c38-e01c-11e7-b2e9-8c636f076c76/3_1513190211004/5b593c38-e01c-11e7-b2e9-8c636f076c76_5a3144dee4b0d6b173bf2785.vtt”,
+          “confidence”: 1.0,
+          “language: “English”,
+          “language_code”: “eng”
```
chadmoone: I think, in general, this is a good addition and makes sense to me. Two questions:

1. Do these `language` and `language_code` property names follow style/convention of the rest of ANS? I know I've seen other examples where the user-facing/pretty/display property has the qualifier (eg `display_date`, etc), and the actual value property may or may not have a qualifier as to the format.  I think there's a lot of value in trying to stick to whatever conventions we've established for things like this. So my vote would be to just prepend `display_` onto the display property here, but if anyone has a better or more definitive answer, please chime in.
2. Further, there is a `locale` trait already in ANS. I know that a "locale" is distinct from (or a sort of a superset of) a "language", and there may be an argument not to use it here—however, this is the trait that we are already using to specify language for content in general, so it would need to justify breaking convention: https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.5.9/content.json#L29.

gengel: I agree with point #1 here, maybe this would be better?
```
{ 
   "language": {
     "language_code": "eng",
     "language_display": "English"
   }
}
```

It's a little clunky and verbose, but one fewer top level property, and more explicit about what the difference is. (That said, I'd also be okay with accepting the PR as is, I don't think there's that much precedent here for blocking this)

For #2, I think the main difference is that `locale` is a property on the document, whereas this is specific to subtitles.  I think the usage is different enough to warrant different fields. (One would expect a video to have one locale with multiple subtitles in specific different languages.) So I'm fine as is.

williamslm: Okay, I'll update.

chadmoone: Sounds good to me 👍

PR #158, opened 2018-01-09
Multisite 0.6.0

gengel:
Adds proposal and basic implementation of multi-site schema


jpblair:
> website - Available only when fetching from Content API for this release
1.  Does this mean that it will be present in the Kafka messages WebSked reads?  We were planning, at least in this first phase, to ingest only documents where `canonical_website == website`.
2.  Are `taxonomy.sections` and `websites` the same thing, but in different formats (array vs. object), or is there a functional difference between them?

gengel:
@jpblair  -- `taxonomy.sections` is queryable and includes the list of all sections a document is in. `websites` is not queryable, can have at most one section per website, and is mainly used for determining the urls

PR #159, opened 2018-01-19
2018-01-19 Proposal for divider content element

epascarello:
Add a content element that allows the author to specify a location where they would like a division in the document.


epascarello:
anyone?

gengel:
Re-opened for broader approval

PR #160, opened 2018-02-06
Proposal: Line Count Planning

brendaneamon:
This PR is for a simple change that will ensure consistency in how story length data are ingested and processed by WebSked.

### DIFF
--- a/docs/proposals/2018-02-06_Line_Count_Planning.md
+++ b/docs/proposals/2018-02-06_Line_Count_Planning.md
@@ -85,19 +85,17 @@ WebSked expects to ingest data in the format below. (Note that the story object 
 
 ```json
 {
-    "_id": "BKZSUVTBGVACFEQDRW7QRXPHFY",
-    "type": "story",
-    "version": "0.6.0",
-    "content_elements": [],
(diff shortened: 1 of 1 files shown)


jpblair:
@brendaneamon In your example, `story_length` is a top-level key, instead of a subobject under `planning`.  I assume this is a typo?

brendaneamon:
@jpblair, yes that was definitely unintentional—I went ahead and made a correction.

gengel:
Are inches normally used as the height metric in non-US newspapers?

jpblair:
I'm not sure, but inches are already part of the schema - what Brendan is looking to add is lines.

RyanGladstone:
@gengel I asked Willi. He said that lines are the most common unit in use. 

He says that some European publishers measure in modules, but isn't sure if that's still a common practice.

A few publishers measure in [cicero](https://en.wikipedia.org/wiki/Cicero_(typography)) or picas https://en.wikipedia.org/wiki/Pica_(typography)).

RyanGladstone:
A representative of Le Parisien said that they use "character count, including spaces" to measure story length.

PR #161, opened 2018-02-21
Distributor fields

gengel:
Adds proposal for distributor fields and owner field re-alignment

### DIFF
--- a/docs/proposals/2018-02-21-Distributor.md
+++ b/docs/proposals/2018-02-21-Distributor.md
@@ -23,7 +23,7 @@ The following fields are to be *deprecated* and ultimately fall into desuetude:
 
 The following fields should be added:
   * `distributor.name` -- The human-readable name of the person or organization controls the distribution rights for a piece of content. E.g. Reuters, AP, Getty. *This replaces some previous uses of `owner.name` and `source.name`.*
-  * `distributor.category` -- The category of distribution method, e.g. Staff/Wire/Feelance/Stock/Handout.  *This replaces some previous uses of `source.type`.*
+  * `distributor.category` -- The category of distribution method, e.g. Staff/Wire/Freelance/Stock/Handout.  *This replaces some previous uses of `source.type`.*
 
 
 ## Example
--- a/docs/proposals/2018-02-21-Distributor.md
+++ b/docs/proposals/2018-02-21-Distributor.md
@@ -24,11 +24,10 @@ The following fields are to be *deprecated* and ultimately fall into desuetude:
 The following fields should be added:
   * `distributor.name` -- The human-readable name of the person or organization controls the distribution rights for a piece of content. E.g. Reuters, AP, Getty. *This replaces some previous uses of `owner.name` and `source.name`.*
   * `distributor.category` -- A machine-readable enumeration of the general category of distribution method: `staff`, `wires`, `other`.  *This replaces some previous uses of `source.type`.*
-  * `distributor.description` -- A human-readable description of the category of distribution method. Particularly used in conjunction with `distributor.category:"other"` but may also be informative in conjunction with `staff` or `wires`.  E.g. "Freelance" or "Staff" *This replaces some previous uses of `source.type`.*
+  * `distributor.subcategory` -- A machine-readable description of the sub-category of distribution method. Used in conjunction with `distributor.category`. E.g. "freelance" or "unsolicited" *This replaces some previous uses of `source.type` and `owner.name`.*
 
 
 ## Example
-
 ```json
 {
   "_id": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
@@ -58,9 +57,12 @@ The following fields should be added:
   },
 
   "distributor": {
-    "name": "Reuters",
-    "category": "wires",
-    "description": "Wires - Reuters"
+    "name": "The Washington Post",
+    "category": "staff",
+    "subcategory": "freelance - unsigned",
+    "additional_properties": {
+      "description": "Something something something"
+    }
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/proposals/2018-02-21-Distributor.md
```diff
@@ -0,0 +1,99 @@
...
+  * `source.edit_url`
+  * `copyright`
+
+The following fields are to be *deprecated* and ultimately fall into desuetude:
+  * `source.name`
+  * `source.type`
+  * `owner.name`
+
+The following fields should be added:
+  * `distributor.name` -- The human-readable name of the person or organization controls the distribution rights for a piece of content. E.g. Reuters, AP, Getty
```
timothykim: Does this means "AP" and "Associated Press" will be treated as two different distributor?

gengel: Yes. Do you think we should add an `id`?

timothykim: I'm not sure. If this field is to categorize a document then maybe? But we can leave it up to the users to maintain the uniqueness of the value.

awong7:
I'm on board with this. Just two points:

- Will need to involve the technical account and delivery managers to communicate this to customers and remind them.
- We should discuss with them what windows of time customers will have in phases two and three since we are putting expectation on customers to schedule the work, get their dev teams on it, and tested in this time frame.

@epascarello @jon-peterson, can you add your thoughts for AF and EL? 

FYI @rowhawn @vgritsenko

jpblair:
Just to codify here the points discussed in the ANS Slack channel:

- The set of values for `distributor.category` here (`Wire`, `Staff`, etc) is inconsistent with the current conventions of `source.source_type`, as followed by WebSked, Ellipsis, Arc I/O, Arc Wordpress and client editors that are integrated with WebSked (`wires`, `staff`).  To change the convention would necessitate migrating a significant amount of data and so would be highly undesirable, at least for WebSked.
- Anglerfish appears to follow yet another convention.  `owner.name` seems to follow the expected convention for `source.source_type`, albeit with a few additional values that are not currently accounted for in other apps (to my knowledge).  `source.source_type`, however, is a free-text field.
- I'm not sure how Goldfish uses these fields, but I believe everything else follows one of the two patterns above.
- @gengel suggested that while we're doing a field migration, we explicitly enumerate the allowed values for `distributor.category` in ANS.  Speaking for WebSked, we support this idea since it enforces consistency and eliminates the likelihood of running into this problem in the future.  The drawbacks are lack of client configurability (i18n?), and the effort of migrating existing data that doesn't conform to the enum.  On the second point, Anglerfish would likely bear most of that effort.  Perhaps the fact that they're already moving fields and the alignment of their current usage of `owner.name` with the enum values would make that easier?
- Greg also proposed that `distributor.category` could be split out into two fields: one that functions like a slug for filtering, conforming to the enum values, and another that is human-readable for display purposes.  I'm a little unclear as to the purpose of the latter, and what current functionality it replaces, but I'll let him speak to that.

gengel:
@jpblair The purpose of the human-readable field is to capture the data that is currently in `source.name` and non-compliant with the heretofore-informal WebSked enum. I will update the proposal to make this more clear

@jon-peterson Can you elaborate more on TGAM's usage of `owner.name`

jon-peterson:
@gengel I asked Adam and Alasdair about tgam and `owner.name`. 

Via the picklist, `owner.name` can take on the values of `staff`, `wires`, `unsolicited`, `freelance signed`, or `freelance unsigned` for tgam. The value of that field controls "a variety of intellectual property logic downstream" in tgam's syndication system and others, and it can change the way the content is displayed on the front end. So changing its representation will definitely require coding work on their part.

It isn't clear why `owner.name` was chosen for this purpose -- there's a tgam ticket (8) that covers moving from `source.name` to `owner.name` but it dates from mid 2016 and includes no rationale.

My two cents: I know migrating will be painful for them and for us, but in my opinion that is less painful than living with the current owner/status confusion for years to come.

willvanwazer:
I like many aspects of this proposal, but what seems weird to me about the example object is the placement of `owner.sponsored`--is there any better place we can put this in the new schema?

gengel:
@willvanwazer I agree that it's not where I'd place it were I starting from scratch, but it's already in place and usage has been consistent.  Changing it feels like forcing people to breaking things for cosmetic purity. The other breaking changes we're making are fixing fields that have, as I understand it, real-world collisions

gengel:
I added `distributor.description` but now I'm not sure that it solves the problem.  It seems like there is a real need for "this field can be anything generally, but within organization x, it is limited to these n values."  This is not easily expressible in one universal set of JSON schema.

gengel:
I've replaced `distributor.description` with `distributor.subcategory` -- a machine-readable *non-enumerated* field that can be used in conjunction with `category` to broadly classify distribution method.

robpc:
I think there is perhaps a typo in the proposal that would like to point out. After the example object, the paragraph begins 

> In this example, the content was ingested via a wire feed from Reuters.

From the Proposal section this seems inconsistent with the example. From that sentence, and the proposal, I would expect `distributor.name` to be `Reuters` and `distributor.category` to be `wires`.

As a follow on question, I am curious if the `owner.id` is also incorrect and should be a "machine-readable" summary of the distributer block (ex `reuters`) or is merely an identifier of the org managing the content system. Said another way, if I was looking to filter out any Reuters content would I be looking at the `owner.id` or the `distributer.name`? If it is the latter, I am concerned that the "human-readable" language in the proposal means that there is not an expectation of some consistency (ie machine entered versus human entered) and would make this type of filtering difficult.

gengel:
Thanks @robpc for pointing out my mistake. I changed the example but forgot to update the description.

`owner.id` is indeed simply the identifier of the organization whose content is being stored

robpc:
> owner.id is indeed simply the identifier of the organization whose content is being stored

Sorry to be a stickler, but that is a bit ambiguous. I was actually hoping you were going to fix the example and not the description as that would be a little more descriptive. In the Reuters example, would the `owner.id` be the same as it is now, or something that identified Reuters?

jpblair:
The overall idea here should work for us.  My only remaining question is whether there are additional values that should be included in the `distributor.category` enum - that could be worth some investigation from product managers.  We can always add to it in the future, though, so it's not a blocker.

PR #166, opened 2018-04-25
Creates ANS 0.6.2 with new category values for distributor.category trait

jodieburnett:
This is the implementation of the proposal in the CA-548-Proposal branch. 

The relevant file is src/main/resources/schema/ans/0.6.2/traits/trait_distributor.json
- "stock" and "handout" were added to the category.enum array
- category.description was updated to include an explanation of when to use these new values


jpblair:
It won't let me review, but LGTM.

jmjakes:
lgtm

PR #167, opened 2018-06-04
Collections

gengel:
Adds proposal and initial schema implementation for new ANS features to support Collections.

* `collection` type

* `promo_reference` (alternative to `reference`)

* `content_aliases` (alternative to `_id`)


docs/proposals/2018-06-04_Collections_Promo_References_Aliases.md
```diff
@@ -0,0 +1,156 @@
...
+ * Promo references only reference story, gallery and video documents.  Additional reference types may be added at a later time.
+
+ * Promo references have the same form as a regular reference in most other respects, but their type is `promo_reference`.
+
+
+### New ANS Content Type: Collections
+
+A new document type `collection` represents in ANS a curated sequence of content. A collection has the same ANS fields as a `story`. However, collections have some significant behavioral differences.
+
+    * `content_elements` in a `collection` may only consist of promo references
```
chadmoone: Is `collection` then not an overly generic term for this? Can we not think of any potential future scenarios where we'd want to represent an ordered collection of items/references other than this specific use case?  I wonder if `promo_collection` (or something along those lines) wouldn't be a better name.

gengel: I think the product name "Collections" is set in stone at this point, but I'll verify

gengel: So it's not set in stone -- but it's the most likely name at this point. 

As for being overly generic, collections are indeed intended to cover a wide variety of use cases, so the name is necessarily generic.  (Lists on homepages, "story galleries", promo boxes, etc)

If we need a different type down the line that is more specific, or that is also somewhat generic, I think we still have "lists", "sets", "sequences", "featured_items_list" etc available to us at that time.

chadmoone: Ok. Personally I still find it overly generic, because I can think of lots of use cases for an ordered collection of items which is not restricted to `promo_item`, and I don’t see the harm in specificity—but I also don’t feel that strongly about it and mostly just wanted to raise and make sure we considered. So if you’re good with it then I’m good with it.

PR #168, opened 2018-07-30
Add scores to Named Entities

fenstamaker:
This as an optional `score` field to the Named Entities in that `taxonomy` section of a story. The scores represent the relevancy of the Named Entity to that story. Without scores, it's impossible to determine if a story is more about one Named Entity over another.

This PR also adds associated tests surrounding the `score` field.


fenstamaker:
Let me fix the schema-tests-07.js file

fenstamaker:
The `schema-tests-07.js` file has been fixed

PR #169, opened 2018-07-31
Content restrictions

gengel:
Adds proposal for forbidden websites property in `content_restrictions`


chadmoone:
Why the blacklist approach? Whitelist seems more intuitive, less error-prone, and arguably a much more accurate representation of the licensing that we’re trying to represent here. Both are going to require migrations in practice.

gengel:
I agree the whitelist is a better match and more intuitive. 

We have millions of documents already without any content restrictions at all, and a half dozen apps that produce content with no whitelist field. So we have a few options for documents without a whitelist:

1) Documents with no whitelist are allowable everywhere.  This is the opposite of how we handle website ids in the publishing `websites` object.  So omitting that latter field would cause the document to disappear everywhere, but omitting the whitelist would do nothing.  At first that seemed unintuitive to me, but maybe if they're named differently enough it could work.

2) Documents with no whitelist are forbidden everywhere. This is basically a breaking change, as all existing documents would have no whitelist and the current production apps don't add it either.  So you could tweak it by adding a feature flag field, like:

```
{ 
  "content_restrictions": {
    "allowed": {
      "enforce_whitelist": true,
      "websites": ["the-post"]
    }
}
```

That's...ok maybe? But it's also one more field to index and one more field to store and these docs are already huge.

3) Another option would be to include both a whitelist and a blacklist, and, as in option 1, do nothing with either if neither field is present.  But in that case you need to make a bunch of decisions about what to do if both or present (e.g. prohibit vs ignore vs override) which will ultimately get passed on to either editor software or reader software.

The one thing that I like about the blacklist option is that it makes every subsequent decision a little easier.

docs/proposals/2018-07-03_allowed_websites.md
```diff
@@ -0,0 +1,100 @@
...
+
+In the above example, "Dog Bites Man" is *available* to be published to The Times and The Post. 
+It is currently published only on The Times. It may never be published on The Herald unless the content restrictions 
+are also changed.
+
+## Questions 
+
+*Who is responsible for enforcing the content restrictions?*
+
+Rendering applications (PB Templates, Apps) will ultimately be responsible for omitting restricted content.
```
timothykim: I'm not sure if this is a good idea. We want least amount of business logic in rendering layer. Since Content API already has business logic to filter out content based on websites, shouldn't Content API own the business logic of enforcing this?

docs/proposals/2018-07-03_allowed_websites.md
```diff
@@ -0,0 +1,100 @@
...
+*Who is responsible for enforcing the content restrictions?*
+
+Rendering applications (PB Templates, Apps) will ultimately be responsible for omitting restricted content.  
+Middleware applications such as Content API may add flags to retrieval endpoints to remove these earlier in 
+the stream if desired. Content production applications may also apply some validation at content save/publish time.
+
+*Who is responsible for populating this value?*
+
+Organizations who wish to take advantage of content restrictions are responsible for populating this value
+and enforcing correctness. Production applications are responsible only for allowing this field to be edited.
```
timothykim: By edited, do you mean created as well?

jpblair:
This only pertains to the rendering layer, right?  Is there an expectation for editorial-facing apps to do anything with this information?  If not, I don't think this will affect WebSked, so I don't have any feedback to provide.

gengel:
@jpblair I think that's a question for @awong7 and @RyanGladstone more than me. But FWIW I would anticipate some kind of label or flag like "This is restricted content" in the editorial apps.

Tronc has a fairly urgent need for this, that's why I'm bringing this up now.

RyanGladstone:
Re: WebSked
At most, WebSked will ingest and display this information. I will wait to hear from clients if want to see this information within WebSked.

PR #172, opened 2018-09-14
Add associated_tasks stanza to taxonomy

jon-peterson:
This is per proposal "task related media", at PR 170. Limit the items to a list of no more than 200 valid WebSked task IDs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "main": "./lib/ans.js",
   "scripts": {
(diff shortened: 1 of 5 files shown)


jodieburnett:
@jon-peterson are we creating a new ANS version along with this change? Should the CAPI devs handle that?
(maybe this is a better question for @gengel)

jon-peterson:
@jodieburnett I thought that the dev version is 0.7.1, so that's where I put it... happy to re-roll this PR if 0.7.2 is better.

gengel:
0.7.1 sounds right to me

PR #173, opened 2018-10-17
CA-1048: referenced distributor proposal

wapo-omar-khan:
https://arcpublishing.atlassian.net/secure/RapidBoard.jspa?rapidView=108&projectKey=CA&modal=detail&selectedIssue=CA-1048

This is to add referencing distributors by id as an option.

### DIFF
--- a/docs/proposals/2018-10-16 - Referenced Distributor.md
+++ b/docs/proposals/2018-10-16 - Referenced Distributor.md
@@ -1,0 +1,40 @@
+# REFERENCED DISTRIBUTOR
+
+# Problem
+
+Users can currently only reference distributors directly. Users need to be able to reference distributors by id.
+
+# Proposal
+
+Add an alternative set of fields to the distributor content element.
+
+## Id
+
+The distributor entry should contain the referenced distributor's id.
+
+## Mode
+
+This enum field will distinguish between whether this field is being used as a referenced or custom distributor.
+
+# Example
+
+```
+{
+  "type": "story",
+  "version": "0.7.2",
+  "distributor": {
+    "id": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
+    "mode": "reference"
+  }
+}
+```
+
+# Concerns
+
+## Will this overlap with the existing distributor schema?
+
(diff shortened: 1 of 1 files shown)


jpblair:
Would the Content API denormalize the distributor reference?  As long as our ingestion process can read the properties we need off of it, then it shouldn't matter if it's a reference.

gengel:
@jpblair Actually I believe the plan was *not* to denormalize this data in so that restrictions can be enabled and disabled across large swaths of content quickly.  @jon-peterson @RyanGladstone This seems like an oversight in the solution design process.  I think there may be WebSked tickets for content restrictions after all. It sounds like it will mainly come down to WebSked accessing the distributor API same as the other editorial apps, but you two should probably confirm that.

RyanGladstone:
@gengel @jpblair I thought that the restriction data was *not* going to be denormalized for the reasons Greg outlines but I expected that the distributor name/category/subcategory would still be available without making an API call. Is that not the case? WebSked doesn't need the website restrictions.

jon-peterson:
@RyanGladstone Greg is correct, references to distributors will not be denormalized; instead, we just maintain the reference in the otherwise-fully-denormalized ANS and contact Distributor API only when we need to. Otherwise we wouldn't be able to guarantee that changes to the restrictions would take effect within the allotted 30 minutes.

RyanGladstone:
https://arcpublishing.atlassian.net/browse/WSK-1673 
**Cache distributor data in the Arc WebSked Pusher**

gengel:
@jpblair Even if we could do the additional denormalization, I don't think that would be correct -- a distributor hard-coded into the content 0.6.x-style might mean something different from a distributor that is referenced in global settings.  Even if this distinction doesn't matter to WebSked, it likely would to other downstream consumers that care about restrictions likely would. (And as a practical reality, we simply cannot do any more denormalization that we already do.) 

But ignoring the implementation for the moment, do you object to this format for the distributor id? I'm happy to continue the implementation conversation elsewhere, but this PR is just meant to reach consensus on how the data is represented.

PR #174, opened 2018-10-24
Add Paragraph Colored Highlight Info Proposal

yury-sannikov:
# Adding Color Metadata to the paragraph

# Problem

Infobae wants to be able to highlight text in paragraphs. An example story is
https://www.infobae.com/opinion/2018/08/08/el-sonido-de-la-sororidad/

Colors are restricted to the text elements only. Hovewer, it might apply later to the other content elements.

# Proposal

Create a new field `colors` in the `story_elements/text.json`. Also, create a trait named `trait_colors`. It should be an array. Each item should contain the following fields:

- `start` of type `number` which contains a color start position in a text
- `length` of type `number` which will be a non-zero value of the colored text
- `class_name` of type `string` which will be a CSS class name applied to the HTML `<span>` element. Its name comes from the Ellipsis settings.

### trait_colors

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_colors.json",
  "title": "Colors",
  "description": "Trait that applies a list of colors to a paragraph.",
  "type": "array",
  "items": {
    "type" : "object",
    "properties" : {
      "start" : {
        "description": "Color start position",
        "type" : "number",
        "minimum": 0
      },
      "length" : {
        "description": "Color length",
        "type" : "number",
        "minimum": 1
      },
      "class_name" : {
        "description": "A CSS class name to the HTML element for the Page Builder",
        "type" : "string",
        "minLength": 1
      }
    },
    "required": [ "start", "length", "class_name" ]
  }
}
```

### story_elements/text.json

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/story_elements/text.json",
  "description": "A textual content element",
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "type": {
      "type": "string",
      "enum": [ "text" ]
    },
    "_id": {
      "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_id.json"
    },
    "subtype": {
      "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_subtype.json"
    },
    "channels": {
      "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_channel.json"
    },
    "alignment": {
      "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_alignment.json"
    },
    "colors": {
      "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_colors.json"
    },
    "additional_properties": {
      "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_additional_properties.json"
    },
    "content": {
      "description": "The text of the paragraph.",
      "type": "string"
    }
  },
  "required": [ "type", "content" ]
}

```

# Concerns

Overlaps and out of range color ranges

Color selections should not overlap and not be out of the text length range. Overlapped color will be applied over the previous color and replace it in the occurrence order. Ranges out of the text length boundaries will be discarded or trimmed to the boundary.

# Implementation

Jon Peterson on the Anglerfish team will implement this and add it to the schema if this proposal is accepted.

### DIFF
--- a/docs/proposals/2018-10-24-Paragraph_Color_Metadata.md
+++ b/docs/proposals/2018-10-24-Paragraph_Color_Metadata.md
@@ -14,7 +14,6 @@ Create a new field `colors` in the `story_elements/text.json`. Also, create a tr
 - `start` of type `number` which contains a color start position in a text
 - `length` of type `number` which will be a non-zero value of the colored text
 - `class_name` of type `string` which will be a CSS class name applied to the HTML `<span>` element. Its name comes from the Ellipsis settings.
-- `color_code` of type `string` having `#RRGGBB` color code. This is for backward compatibility, if Ellipsis has no color settings, it will still be able to render colored text. It also might be used by Page Builder to generate styles on the fly or even inject a text with inline styles, if needed.
 
 ### trait_colors
 
@@ -42,14 +41,9 @@ Create a new field `colors` in the `story_elements/text.json`. Also, create a tr
         "description": "A CSS class name to the HTML element for the Page Builder",
         "type" : "string",
         "minLength": 1
-      },
-      "color_code" : {
-        "description": "A HTML color code for the text block",
-        "type" : "string",
-        "pattern": "#[0-9a-fA-F]{6}"
-      },
+      }
     },
-    "required": [ "start", "length", "class_name", "color_code" ]
+    "required": [ "start", "length", "class_name" ]
   }
 }
 ```
--- a/docs/proposals/2018-10-24-Paragraph_Color_Metadata.md
+++ b/docs/proposals/2018-10-24-Paragraph_Color_Metadata.md
@@ -1,0 +1,107 @@
+# Adding Color Metadata to the paragraph
+
+# Problem
+
+Infobae wants to be able to highlight text in paragraphs. An example story is
+https://www.infobae.com/opinion/2018/08/08/el-sonido-de-la-sororidad/
+
+Colors are restricted to the text elements only. Hovewer, it might apply later to the other content elements.
+
+# Proposal
+
+Create a new field `colors` in the `story_elements/text.json`. Also, create a trait named `trait_colors`. It should be an array. Each item should contain the following fields:
+
+- `start` of type `number` which contains a color start position in a text
+- `length` of type `number` which will be a non-zero value of the colored text
+- `class_name` of type `string` which will be a CSS class name applied to the HTML `<span>` element. Its name comes from the Ellipsis settings.
+- `color_code` of type `string` having `#RRGGBB` color code. This is for backward compatibility, if Ellipsis has no color settings, it will still be able to render colored text. It also might be used by Page Builder to generate styles on the fly or even inject a text with inline styles, if needed.
+
+### trait_colors
+
+```
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_colors.json",
+  "title": "Colors",
+  "description": "Trait that applies a list of colors to a paragraph.",
+  "type": "array",
+  "items": {
+    "type" : "object",
+    "properties" : {
+      "start" : {
+        "description": "Color start position",
+        "type" : "number",
+        "minimum": 0
+      },
+      "length" : {
+        "description": "Color length",
+        "type" : "number",
+        "minimum": 1
+      },
+      "class_name" : {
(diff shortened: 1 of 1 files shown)


RyanGladstone:
Would it be hacky for a client to use this feature to change the text color instead of adding a highlight?

Beyond the PR title, I don't see anything that's particularly highlighting-focused.

The Washington Post is interested in a feature that'll allow them to highlight characters in red, as seen here: https://www.washingtonpost.com/news/style/wp/2018/10/21/evan-birnholzs-oct-21-post-magazine-crossword-political-ascension/?utm_term=.f812563bf5aa

gengel:
@RyanGladstone I think the Post's use case is perfectly in line with this. That said, I am glad that @yury-sannikov removed the exact color value and simply used class name. I would probably change the field names to "highlight" or something, however, rather than color -- it's easy to imagine sites rendering "highlighted" text with mechanisms other than (or in addition to) color changes.

One more question -- is the intention here that the `class_name` is selected from a picklist configured in Ellipsis or free-text?

tbrick855:
Couldn't this be made more general purpose with a tag and attributes.  Ellipsis is already doing something like this in inline_comments using pos and comment.  Seems like this approach will require adding a new field for each new tag type.

chenghsinw:
what's the main reason we can not treat highlight text like bold, italic and underline these kind of text styles by using span tag with a specific class?

jpblair:
I don't think WebSked would do anything with this color info, so not sure it has any effect on us since it appears to be an additive change.  I think Ryan is correct that there's some ambiguity as to whether this refers to text color or background/highlight color, but I don't think WebSked has any skin in the game.

chadmoone:
Hey guys—I’m on vacation until Monday and will have very limited time to review this further and respond in more detail this week, but I’d like to make clear that I’m staunchly opposed to doing this and beg you to hold until next week. The solution to this problem is very clearly https://github.com/washingtonpost/ans-schema/pull/134, and if we had not delayed that continually, this proposal would not exist. Introducing this on its own as proposed will only make the real solution more convoluted and probably delay it more. It’s clear we have a need for an attributes system. Can we please accelerate that instead?

Additionally I might have qualms with the HTML-centricity of `class_name`...

yury-sannikov:
> Couldn't this be made more general purpose with a tag and attributes. Ellipsis is already doing something like this in inline_comments using pos and comment. Seems like this approach will require adding a new field for each new tag type.

It's already inherited inline comments structure, except it does not contain text. That's because inline comments should not be present in the text, but highlights should.

yury-sannikov:
@gengel 

> I would probably change the field names to "highlight" or something, however, rather than color -- it's easy to imagine sites rendering "highlighted" text with mechanisms other than (or in addition to) color changes.

Make sense to me.

> One more question -- is the intention here that the `class_name` is selected from a picklist configured in Ellipsis or free-text?

That's the details we are trying to figure out with Angela. I suppose, it will be a free-text list, added by Ellipsis administrator according to the requirements. Also I have some doubts how it should support multi site set up. Different sites might involve different highlight classes. Some of those classes migh need to be ignored for certain sites.
That way, `class_name` might become an object, where `key` is a site ID and value is a class name. For some sites it might be undefined, meaning no highlight should be applied.

yury-sannikov:
> what's the main reason we can not treat highlight text like bold, italic and underline these kind of text styles by using span tag with a specific class?

@chenghsinw 

Looks like its's not possible with Draft HTML->Ans convertor to process conversion based on a span custom styles. It can be done through custom block types and then merged, but it's hacky.

Also ANS initiative is to get rid of any HTML markup (pita for mobile teams, etc) and store metadata along with the plain text.

yury-sannikov:
> #134

@chadmoone 

That actually has more sense to me and will solve it in a right way. However, Im not sure if it feasible to roll it out in the near future. Let me bring this up with @RyanGladstone @gengel

yury-sannikov:
Closing, since we came up with a different solution. Thanks, everyone.

PR #178, opened 2018-11-01
Alternate Credits Proposal

gengel:


docs/proposals/2018-11-01_-Alt_Credits.md
```diff
@@ -0,0 +1,89 @@
...
+      "credits": {
+        "by": [
+          {
+            "type": "author",
+            "version": "0.8.0",
+            "byline": "Jim Phelps"
+          }
+        ]
+      },  
+      "alternate_credits": {
```
chadmoone: How does this work if we have multiple alternate credits?

docs/proposals/2018-11-01_-Alt_Credits.md
```diff
@@ -0,0 +1,89 @@
...
+        "by": [
+          {
+            "type": "author",
+            "version": "0.8.0",
+            "byline": "If you do not wish to display the actual photographer, you can display this message instead."
```
chenghsinw: Can users use empty string here so nothing to display?

docs/proposals/2018-11-01_-Alt_Credits.md
gengel: Do you mean, if there are alternates for more than one context? (E.g., show this on iPhone, this on iPad, this on web, this on the feed...) This proposal does not support that.

docs/proposals/2018-11-01_-Alt_Credits.md
gengel: The ANS preference historically for multiples has been an array of 0 elements. E.g.:

```
  "alternate_credits": {
    "by": []
  }
```

However, we should stay consistent with how we're storing data in `credits`.  What does Ellipsis do in the case that a user does not specify a credit in Ellipsis?

docs/proposals/2018-11-01_-Alt_Credits.md
chadmoone: Right. Or another example would be www, rainbow, apple news, the lily. And this is a pretty real scenario, though most of those are not using ANS directly yet.

I think this is too fragile. If we want to support multiple bylines, we should be supporting for `n`.

docs/proposals/2018-11-01_-Alt_Credits.md
chenghsinw: Okay, for stories it's an empty array if users don't specify in ellipsis, so I think we can use the same approach to stay consistently.

docs/proposals/2018-11-01_-Alt_Credits.md
gengel: If you're talking about the rainbow use case, I think that needs a much, much broader solution -- potentially any field needs an alternate, over some (hopefully enumerated) set of contexts.  I'd rather solve that with generic overrides or story branching than individual "alternate" fields.  This is intended as a point solution for a very specific use case -- wanting to vary the "nice presentation" of certain embedded images from the "real" version that we send downstream to consumers.

One point that this does raise is that "alternate" is too vague a name -- we should probably be explicit about exactly where this set of credits should be used instead of the regular set.

@chadmoone Does this answer alleviate your concerns at all?

anscho: > One point that this does raise is that "alternate" is too vague a name -- we should probably be explicit about exactly where this set of credits should be used instead of the regular set.

This is a good point. I don't know if the focus is more on "vanity" or "short," but I appreciate precision of specification.

gengel: Updated proposal to better describe the feature, and change the name to `preferred_credits`. I'm not sure it's all *that* much better.

Other options:

- `native_credits`
- `local_credits`
- `render_credits`
- `reader_credits`
- `curated_credits`

I hate all of these.

awong7:
oops i meant this one @chadmoone. do you have outstanding questions or concerns?
also we're waiting review from @anscho when he is freed up from elections

docs/proposals/2018-11-01_-Alt_Credits.md
chadmoone: Alright, I guess (as mentioned in slack) I'd just prefer a more precise name if possible. Of course naming things is hard, so whatever we can get consensus around is fine.

PR #180, opened 2018-11-13
Adding raw_html type into promo_items proposal

yury-sannikov:
# Adding raw_html type into promo_items

# Problem

Tronc wants to set an arbitrary raw HTML as a promo item. This should include `basic` key in the `promo_items` object. Currently, only [reference](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.7.1/utils/reference.json) and [content](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.7.1/content.json) types are allowed under `promo_items`.

# Proposal

Add [raw_html](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.7.1/story_elements/raw_html.json) as a supported type to the both `basic` and all other keys under `promo_items` object.

### trait_promo_items

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_promo_items.json",
  "title": "Promo Items",
  "type": "object",
  "description": "Lists of promotional content to use when highlighting the story. In the Arc ecosystem, references in these lists will be denormalized.",
  "properties": {
    "basic": {
      "oneOf": [
        {
          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/content.json"
        },
        {
          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/utils/reference.json"
        },
        {
          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/story_elements/raw_html.json"
        }
      ]
    }
  },
  "patternProperties": {
    ".*": {
      "oneOf": [
        {
          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/content.json"
        },
        {
          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/utils/reference.json"
        },
        {
          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/story_elements/raw_html.json"
        }
      ]
    }
  }
}
```

### Sample Story
```
{
  "type": "story",
  "version": "0.8.1",
  "owner": {
    "id": "staging"
  },
  "promo_items": {
    "basic": {
      "type": "raw_html",
      "content": "<h1>raw HTML content</h1>"
    }
  },
  "content_elements": [
    ...
  ],
  "credits": {
    ...
  },
  "taxonomy": {
    ...
  }
}
```

# Concerns

A `basic` key has its own set of validation rules. This might be done for a reason. By default Ellipsis treat `basic` key as an `image` type, however, it's flexible.

# Implementation

Yury Sannikov from Ellipsis team will implement this and add it to the schema if this proposal is accepted.

### DIFF
--- a/docs/proposals/2018-11-13-Raw-HTML-PromoItem.md
+++ b/docs/proposals/2018-11-13-Raw-HTML-PromoItem.md
@@ -1,0 +1,59 @@
+# Adding raw_html type into promo_items
+
+# Problem
+
+Tronc wants to set an arbitrary raw HTML as a promo item. This should include `basic` key in the `promo_items` object. Currently, only [reference](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.7.1/utils/reference.json) and [content](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.7.1/content.json) types are allowed under `promo_items`.
+
+# Proposal
+
+Add [raw_html](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.7.1/story_elements/raw_html.json) as a supported type to the both `basic` and all other keys under `promo_items` object.
+
+### trait_promo_items
+
+```
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/traits/trait_promo_items.json",
+  "title": "Promo Items",
+  "type": "object",
+  "description": "Lists of promotional content to use when highlighting the story. In the Arc ecosystem, references in these lists will be denormalized.",
+  "properties": {
+    "basic": {
+      "oneOf": [
+        {
+          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/content.json"
+        },
+        {
+          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/utils/reference.json"
+        },
+        {
+          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.7.1/story_elements/raw_html.json"
+        }
+      ]
+    }
+  },
+  "patternProperties": {
+    ".*": {
+      "oneOf": [
+        {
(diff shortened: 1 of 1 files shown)


cselvaraj:
Instead just handling "[raw_html](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.8.0/story_elements/raw_html.json)," maybe it should be "[content_element](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.8.0/utils/content_element.json)."  (I think this would require switching to "anyOf" instead of "oneOf" to be backwards compatible.)

That is extremely permissive (only requires the "type" property,) but that will get everything in [story_elements](https://github.com/washingtonpost/ans-schema/tree/master/src/main/resources/schema/ans/0.8.0/story_elements) now (along with some [utils](https://github.com/washingtonpost/ans-schema/tree/master/src/main/resources/schema/ans/0.8.0/utils)), instead of slowly building up to them after the 13 additional proposals are submitted.

This occurred to me for "[related_content](https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.8.0/traits/trait_related_content.json)," since there's a client that has arbitrary links in the related content items of their source data.  ANS was not able to represent a lossless conversion of their data.

There are a few examples in the ANS schema where either a content item or a reference can be placed.  If each of these places also allowed "content_element," that would handle a wide range of situations.

As a slight aside, it might be good to refactor all of those content-or-reference specifications into a single trait.  Had that been the case prior to this proposal, this change could have been placed there and would be applicable in all of the corresponding schemas, like "related_content."

gengel:
@cselvaraj That is all good feedback but probably out of scope for this particular proposal.

In general, we err towards restrictive rather than expressive fields at this point because it is extremely difficult to build applications that can both handle highly-expressive schemas while maintaining decent usability for non-technical users.  

"As a slight aside, it might be good to refactor all of those content-or-reference specifications into a single trait."

This seems correct, but I'm still not sure I'd add _raw_html_ to any of those places.  I *barely* think it's a good idea to allow it in _promo_items_.

PR #184, opened 2018-11-27
EL-2248 propose new html_meta trait

chenghsinw:
Proposing to add a new trait `robot_instructions` for configuring the meta tag instructions for each robot.

### DIFF
--- a/docs/proposals/2018-11-27-html_meta.md
+++ b/docs/proposals/2018-11-27-html_meta.md
@@ -1,4 +1,4 @@
-# Adding robot_instructions
+# Adding html_meta
 
 # Problem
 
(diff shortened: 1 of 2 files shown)


jpblair:
Looks like this is an additive change that WebSked won't use, so no opinion here.

jon-peterson:
I second @gengel 's idea to make this more general; I like the idea of `html_meta`, where each element is the name and content for a specific `meta` tag in the head.

chenghsinw:
@gengel and @jon-peterson I have changed it to html_meta for more generic usage.

chadmoone:
Not to make this too complicated, but now that we're using `html_meta` (which I definitely agree with), should we be supporting all attributes, rather than just `name` & `content`? I don't know how often `charset` or `httpEquiv` are used in practice (`scheme` is deprecated I believe), and possibly this isn't worth taking into consideration, but it seems like if they ever might come into use in the future, we're better off making this map properly now.

chenghsinw:
@chadmoone for charset, it's probably no much need to make it configurable as unicode (UTF-8) is usually used.
I'm not sure about if we should open up http-equiv. @gengel or @jon-peterson do you have any thoughts about it?

chadmoone:
Ok, sure. I'll defer to those who will use this in practice (HTML meta doesn't mean much in native apps world), but just wanted to make sure we considered it.

anscho:
No objections.

willvanwazer:
For implementations we’ve done with clients, we’ve put this inside of the tracking object. Perhaps instead of a new top level property, we define `html_meta` inside of `tracking`?

chenghsinw:
@styonsk could u add me to the repo? I can't access it. Thank you.

styonsk:
Yep! Done

chenghsinw:
This is actually the file https://github.com/wapopartners/partner-tronc-adapter/blob/master/src/troncadapter/adapter/converter/base.py#L175 what the tracking is used

gengel:
@willvanwazer @styonsk I believe the `tracking` field is not generally safe for the platform to use.  @jmjakes can confirm. How bad would sticking with `html_meta` be for PS?

chenghsinw:
Close this PR since using `syndication` is decided as the preferable for the SEO purpose.
We will create a separated proposal to update `syndication` to make sure it's cleared defined and include out use cases.

PR #188, opened 2019-01-08
EL-2342 embargo schema implementation

chenghsinw:
This is the implementation of the embargo proposal https://github.com/washingtonpost/ans-schema/pull/185

### DIFF
--- a/src/main/resources/schema/ans/0.8.1/traits/trait_content_restrictions.json
+++ b/src/main/resources/schema/ans/0.8.1/traits/trait_content_restrictions.json
@@ -8,6 +8,27 @@
     "content_code":{
       "type": "string",
       "description": "The content restriction code/level/flag associated with the ANS object"
+    },
+    "embargo": {
+      "description": "Embargo configuration to enforce publishing restrictions",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "active": {
+          "type": "boolean",
+          "description": "The boolean flag to indicate if the embargo is active or not"
+        },
+        "end_time": {
+          "type": "string",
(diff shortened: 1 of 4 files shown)


chadmoone:
I'll defer to you all on implementation.

src/main/resources/schema/ans/0.8.1/traits/trait_content_restrictions.json
```diff
@@ -8,6 +8,27 @@
...
       "description": "The content restriction code/level/flag associated with the ANS object"
+    },
+    "embargo": {
+      "description": "Embargo configuration to enforce publishing restrictions",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "active": {
+          "type": "boolean",
+          "description": "The boolean flag to indicate if the embargo is active or not"
```
anscho: Can we change this description to indicate that it is an "enable" toggle and not a calculated field? Something like "If this field is false, ignore the embargo." I'd even be in favor of renaming to `disabled` (and not require it), but updating the description would go a long way.

src/main/resources/schema/ans/0.8.1/traits/trait_content_restrictions.json
```diff
@@ -8,6 +8,27 @@
     "content_code":{
       "type": "string",
       "description": "The content restriction code/level/flag associated with the ANS object"
+    },
+    "embargo": {
+      "description": "Embargo configuration to enforce publishing restrictions",
```
anscho: I don't know the "voice" of the ANS schema, but I might write it `Embargoed content must not go live`. No strong opinion here.

PR #189, opened 2019-01-09
AN-1846: Alt Text Proposal

christine-sm:
https://arcpublishing.atlassian.net/browse/AN-1846

This PR includes a proposal for adding `alt_text` to images, as requested by multiple customers. An implementation PR will follow upon approval of this proposal.

### DIFF
--- a/docs/proposals/2019-01-09-Image-Alt-Text-Proposal.md
+++ b/docs/proposals/2019-01-09-Image-Alt-Text-Proposal.md
@@ -2,80 +2,20 @@
 
 # Problem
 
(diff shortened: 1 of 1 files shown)


cmmovius:
@chadmoone thanks for bringing up that point about `description`. It doesn't seem like we surface `description` in the Anglerfish edit image UI, but rather `caption`. However, `description` is used in the edit gallery and edit lightbox view to the same effect as how `caption` is used in the edit image view. As a result, I would still be in favor of having `alt_text`, because its name and purpose is very clear, and would reduce confusion with how `description` is used for galleries and lightboxes.

RyanGladstone:
@jpblair No, WebSked will not be impacted by this PR

PR #190, opened 2019-01-16
AN-1846: implementation of Alt Text proposal

cmmovius:
Alt Text implementation of proposal from PR #189

### DIFF
--- a/src/main/resources/schema/ans/0.8.1/traits/trait_alt_text.json
+++ b/src/main/resources/schema/ans/0.8.1/traits/trait_alt_text.json
@@ -2,6 +2,6 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.8.1/traits/trait_alt_text.json",
   "title": "Alt Text",
-  "description": "A phrase to describe the contents of an image to users.",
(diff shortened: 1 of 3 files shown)


src/main/resources/schema/ans/0.8.1/traits/trait_alt_text.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.8.1/traits/trait_alt_text.json",
+  "title": "Alt Text",
+  "description": "A phrase to describe the contents of an image to users.",
```
chadmoone: Sorry I didn't notice this during the proposal portion, but should we also specify/clarify here that this is used for accessibility?

gengel: Would it be sufficient to say this is the direct ANS equivalent of the HTML `alt` attribute?
https://html.spec.whatwg.org/#attr-img-alt

cmmovius: Good suggestions. I will update this to description to make it more clear that it is for accessibility and that it is "the direct ANS equivalent of the HTML alt attribute".

cmmovius: @chadmoone @gengel I have updated the description accordingly

PR #191, opened 2019-01-16
Proposal to add host_talent and contributors to credit trait

williamslm:
Proposal to add host_talent and contributors to credit trait


williamslm:
> An alternative would be adding one `contributors` array with optional `type` and `description` fields instead of separate key-array pairs.
> 
> 1. That leaves it up to an organization to provide detailed or sparse information on contributors.
> 2. If we continue adding types, systems searching for a particular author won't need to add more locations to search.
> 
> Are there requirements that favor the proposed specification over this suggestion?

gengel:
@anscho The author objects in that array are typically defined in the author service and denormalized in to the body, so the `type` and `description` you suggest would have to be universal to work. I don't think it is the case that we can guarantee that anyone who is `by` would never be a `contributor` or vice versa.

Also, my understanding is that classifying separate credits by type is easier on the rendering side than iterating through a list and separating them out, but I'll defer to @styonsk and @chadmoone on that point

vviswa:
@gengel I agree with your comment above. There would be users who could be in by/host_talent/contributor for the the same Video. We have already implemented the PB features for WaPo to iterate through these individual arrays as you can see from example - https://api.washpost.arcpublishing.com/content/v3/videos?_id=d537c9bc-091d-4cf3-af47-9cc0c3f110c0. (We can change this UX approach - but was cleaner especially if a user in in multiple of these credits fields.)The ANS proposal is to formalize this and have these be indexed by ElasticSearch.

chadmoone:
From my perspective, sure, it's probably a little simpler to ingest if they are broken into separate arrays, but it's not a big deal to parse a single array of `credit` objects.  I also don't see any reason why @anscho's suggestion wouldn't allow a contributor to have multiple credits:

```json
"credits": [
  {
    "type": "author",
    "name": "Mary Beth Albright",
    "contribution_type": "host_talent"
  },
  {
    "type": "author",
    "name": "Linda Williams",
    "org": "The Washington Post",
    "contribution_type": "by"
  },
  {
    "type": "author",
    "name": "Billy Tucker",
    "contribution_type": "contributor"
  },
  {
    "type": "author",
    "name": "Mary Beth Albright",
    "contribution_type": "contributor"
  },
  {
    "type": "author",
    "name": "Linda Williams",
    "contribution_type": "contributor"
  }
]
```

While I do think this makes it a little easier on ANS to add contribution types, many (but probably not all) clients would still need to be updated to understand and properly render the new types, regardless of which way the JSON is structured.

So, in short, from the apps perspective, I see no real concern with either option here, and would defer to the impact they each have on other systems.

jpblair:
@RyanGladstone Do we want these additional roles to be included in the WebSked version (and therefore returned by the author search)?

PR #197, opened 2019-02-08
Author type

gengel:
A proposal to add the `author_type` field to author objects.


st421:
I know author service is special, but having that service conform to ANS as much as possible rather than the other way around (as well as reducing the number of field names everyone has to keep track of) seems preferable. Could `subtype` be used for this?

gengel:
@styonsk @st421 @anscho @jmjakes @williamslm 

Updated to make "author type" map to "subtype."

Can we get this approved and closed out?

PR #198, opened 2019-02-11
Proposal recommended view

robpc:
Proposal for a new field to indicate optimal rendering method

### DIFF
--- a/docs/proposals/2019-02-08 - Recommended View.md
+++ b/docs/proposals/2019-02-08 - Recommended View.md
@@ -1,0 +1,46 @@
+# Proposal for a new field to indicate optimal rendering method
+
+# Problem
+
+Some ANS documents have an critical content to the understanding of the story contained within content elements such as `raw_html` that allow for content that is not well defined by the ANS format. The unpredicability of `raw_html` makes rendering of this content unstable or impossible to some downstream rendering applications (ex. Mobile). In other cases, other rendering applications may have additional information that is not supplied to ANS. Two examples of the latter:
+  - Static graphics pages where the ANS version is the fallback
+  - A story that uses AR in `raw_html` that would be better served on a device
+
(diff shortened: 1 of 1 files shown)


docs/proposals/2019-02-08 - Recommended View.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Alternatives Considered
+
+## Why not simply not render any content that has the problematic `raw_html`?
+
+We already make decisions about whether an article is able to be viewed in our applications, but given the problem it is not always easy to tell. Often we or the content creators want a way to override the default behavior.
+
+## Why not just have a flag to force a webview?
+
+That would work for our immediate needs too, but when thinking about that option, I felt like it was serving a highly specific need. The examples above are intended to show how this concept might apply to other cases, and this was an effort to make a more useful solution.
```
anscho: If the only two views are `www` and `native-app`, I might prefer solving the specific problem (ex: `"relies_on_raw_html": true` or `raw_html_is_central` or whatever).

Either way, I agree that this was inevitable once `raw_html` was included.

robpc: As stated in the part you quoted, open to that too. I think this proposal offers a little more flexibility. The AR example is meant to convey a scenario that the flag doesn't cover. But if collectively that feels like over-engineering the problem I am happy to drop it. I am not a fan of your names, though, and I think they hide the true intent which is to force a webview if available. The name `webview_prefered` is more honest to the intent, though I don't like it either and think is at odds with the rendering agnostic philosophy. That is what started me on this more flexible version which might be more palatable for the purity of spec.

gengel:
@robpc Yeah, agree this is or something very similar is necessary.

Minor aesthetic quibble: is there an existing trait that would be appropriate to stick this in, or else, could we make a new one? The top-level fields on ANS are getting a little crowded; it would be nice to group related things as appropriate to avoid ending up like the six different `something_something_date` fields

robpc:
@gengel I didn't see anything the first time I looked, but will take a note to take another pass.

PR #199, opened 2019-02-11
Added Alternatives to Raw HTML Proposal

robpc:
Proposal to add `alternative` trait to `raw_html` proposal

### DIFF
--- a/docs/proposals/2019-02-08 - Fallback Content Elements.md
+++ b/docs/proposals/2019-02-08 - Fallback Content Elements.md
@@ -6,7 +6,7 @@ The content element `raw_html` is often used to inject Javascript into an ANS do
 
 # Proposal
 
-Add a required `alternative` trait to the `raw_html` content element, that contains an array of one content element (allows for multiple later) that can notionally replace the `raw_html` block. The content in the array can only be of type `text`, `image`, `gallery` or `video`. When `upvert`ing this trait should be filled this with a `text` element that has a stock `This content is not available on this platform` message (or similar).
+Add an `alternative` trait to the `raw_html` content element, that contains an array of one content elements that can notionally replace the `raw_html` block. The content in the array can only be of type `text`, `image`, `gallery` or `video`. This trait is required unless the `rendering_guides.preferred_method` is set to `website` only. In all other cases, when `upvert`ing this trait should be filled this with a `text` element that has a stock `This content is not available on this platform` message (or similar).
 
 ## Alternatives
 
@@ -70,12 +70,6 @@ The social team is trying to build awareness of the Twitch channel so they are e
 Yes, honestly, in my opinion that is the "right" way to handle `raw_html`. The `raw_html` element should ONLY be allowed as an alternative to "pure" ANS content elements and in a well-known named one (ex `browser`) to help rendering applications make the right decision on content. The `raw_html` element is an escape hatch to ANS and defeats many of the benefits of ANS to the rendering layer. There should never be `raw_html` at the root level of the content element list. What I am calling the `basic` alternative above should be the content element or collection in an `element_group` and the `raw_html` could be in the alternatives for those applications that can use it. 
 
 Realistically though, given momentum and the frequent use of raw_html now, I am not sure that this answer would be tenable, so the proposal is a version of this that is more backwards compatible.
-
-## Does there need to be multiple alternatives by name? Couldn't it just be one list?
-
-_EDIT: Changed to this suggestion based on feedback_
-
-Yes, that would definitely solve the issue we are having, but my thinking was that the naming would be a signal as to the capabilities desired by these alternatives. In the proposal `basic` here suggests that you only need to be able to support ANS elements to render this alternative. In the example in the previous question `browser` signifies that it should be rendered in a browser-like environment.
 
 ## Couldn't you use channels instead?
 
--- a/docs/proposals/2019-02-08 - Fallback Content Elements.md
+++ b/docs/proposals/2019-02-08 - Fallback Content Elements.md
@@ -6,24 +6,22 @@ The content element `raw_html` is often used to inject Javascript into an ANS do
 
 # Proposal
 
-Add an `alternatives` trait to the `raw_html` content element, that contains one or more arrays of content elements that can notionally replace the `raw_html` block. The `basic` one is the only one required, and cannot contain a `raw_html` content element. `upvert`ing should fill this with a stock `text` element that has a stock `This content is not available on this platform` message (or similar).
+Add a required `alternative` trait to the `raw_html` content element, that contains an array of one content element (allows for multiple later) that can notionally replace the `raw_html` block. The content in the array can only be of type `text`, `image`, `gallery` or `video`. When `upvert`ing this trait should be filled this with a `text` element that has a stock `This content is not available on this platform` message (or similar).
 
 ## Alternatives
 
-Each `raw_html` element should have an `alternatives` and at least a `basic` array with non-`raw_html` content. An example of the default behaviour:
+Each `raw_html` element should have an `alternative` and at least a `basic` array with non-`raw_html` content. An example of the default behaviour:
 
 ```
 {
   "type": "raw_html",
   "content": "<script src=\"https://www.washingtonpost.com/stat/cool.js\"></script>",
-  "alternatives": {
-    "basic": [
-      {
-        "type": "text",
-        "content": "<b>This content is unavailable on this platform</b>"
-      }
-    ]
-  }
+  "alternative": [
+    {
+      "type": "text",
+      "content": "<b>This content is unavailable on this platform</b>"
+    }
+  ]
 }
 ```
 
@@ -35,18 +33,16 @@ The graphics team has created a JS rendered graphic that scales to browser viewp
 {
   "type": "raw_html",
   "content": "<div class=\"graphics\"><!-- Graphics HTML and JS --></div>",
-  "alternatives": {
-    "basic": [
-      {
-        "type": "image",
-        "url": "https://www.coolpublishing.com/graphics/alt-cool-1.png"
-      }
-    ]
-  }
+  "alternative": [
+    {
+      "type": "image",
+      "url": "https://www.coolpublishing.com/graphics/alt-cool-1.png"
+    }
+  ]
 }
 ```
 
-# Example 2
+# Example 2 (future implementation of multiple elements)
 
 The social team is trying to build awareness of the Twitch channel so they are embedding the Twitch stream where normally we would use the Goldfish live video. They provide the normal video as an alternative as well as a note to encourage users to open the twitch channel.
 
@@ -54,18 +50,16 @@ The social team is trying to build awareness of the Twitch channel so they are e
 {
   "type": "raw_html",
   "content": "<iframe src=\"https://player.twitch.tv/?channel=coolpublishing\" frameborder=\"0\" allowfullscreen=\"true\" scrolling=\"no\" height=\"378\" width=\"620\"></iframe><a href=\"https://www.twitch.tv/coolpublishing?tt_content=text_link&tt_medium=live_embed\" style=\"padding:2px 0px 4px; display:block; width:345px; font-weight:normal; font-size:10px; text-decoration:underline;\">Watch live video from coolpublishing on www.twitch.tv</a>",
-  "alternatives": {
-    "basic": [
-      {
-        "type": "video",
-        // PostTV, etc ...
-      },
-      {
-        "type": "text",
-        "content": "<b>We are also streaming live on <a href=\"https://www.twitch.tv/coolpublishing\">Twitch</a></b>"
-      }
-    ],
-  }
+  "alternative": [
+    {
+      "type": "video",
+      // PostTV, etc ...
+    },
+    {
+      "type": "text",
+      "content": "<b>We are also streaming live on <a href=\"https://www.twitch.tv/coolpublishing\">Twitch</a></b>"
+    }
+  ]
 }
 ```
 
@@ -78,6 +72,8 @@ Yes, honestly, in my opinion that is the "right" way to handle `raw_html`. The `
 Realistically though, given momentum and the frequent use of raw_html now, I am not sure that this answer would be tenable, so the proposal is a version of this that is more backwards compatible.
 
 ## Does there need to be multiple alternatives by name? Couldn't it just be one list?
+
+_EDIT: Changed to this suggestion based on feedback_
 
 Yes, that would definitely solve the issue we are having, but my thinking was that the naming would be a signal as to the capabilities desired by these alternatives. In the proposal `basic` here suggests that you only need to be able to support ANS elements to render this alternative. In the example in the previous question `browser` signifies that it should be rendered in a browser-like environment.
 
--- a/docs/proposals/2019-02-08 - Fallback Content Elements.md
+++ b/docs/proposals/2019-02-08 - Fallback Content Elements.md
@@ -6,7 +6,7 @@ The content element `raw_html` is often used to inject Javascript into an ANS do
 
 # Proposal
 
-Add an `alternative` trait to the `raw_html` content element, that contains an array of one content elements that can notionally replace the `raw_html` block. The content in the array can only be of type `text`, `image`, `gallery` or `video`. This trait is required unless the `rendering_guides.preferred_method` is set to `website` only. In all other cases, when `upvert`ing this trait should be filled this with a `text` element that has a stock `This content is not available on this platform` message (or similar).
+Add an `alternative` trait to the `raw_html` content element, that contains an array of one content elements that can notionally replace the `raw_html` block. The content in the array can only be of type `text`, `image`, `gallery` or `video`.
 
 ## Alternatives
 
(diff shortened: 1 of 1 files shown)


docs/proposals/2019-02-08 - Fallback Content Elements.md
```diff
@@ -0,0 +1,94 @@
...
+        "content": "<b>We are also streaming live on <a href=\"https://www.twitch.tv/coolpublishing\">Twitch</a></b>"
+      }
+    ],
+  }
+}
+```
+
+# Concerns
+
+## Wouldn't this be better flipped? Where the `raw_html` was available as an alternative to an content_element?
```
anscho: I agree that `raw_html` should require a fallback.

gengel:
@robpc What is the allowable set of elements within "alternatives"? Presumably it would not include a nested "raw_html"?

Is the key-value pair necessary? What is the advantage to having multiple named variant alternatives as opposed to a single, safe-to-render-everywhere "alternative?"

docs/proposals/2019-02-08 - Fallback Content Elements.md
```diff
@@ -0,0 +1,94 @@
...
+# Example 1
+
+The graphics team has created a JS rendered graphic that scales to browser viewport size. They also generate static `.png`s as a fallback for mobile and tablet.
+
+```
+{
+  "type": "raw_html",
+  "content": "<div class=\"graphics\"><!-- Graphics HTML and JS --></div>",
+  "alternatives": {
+    "basic": [
```
chenghsinw: what are the cases we will use other keys?
Since it's a fallback, will it be simpler we just have one alternative?

robpc:
@gengel Thanks for the comment. I mention the part about that briefly in the proposal section, but open to better ways to write that. Basically an array of content elements that cannot contain a `raw_html` type. The named is an attempt to allow for different capabilities or targeted alternatives, but open to this being just an array too if the group thinks the proposal is too open ended.

docs/proposals/2019-02-08 - Fallback Content Elements.md
robpc: The intention was to leave it open to other things like AR or some other platform specific content. Less so as a fallback and more as a alternative set of content if the platform supported it. I guess I was oscillating between the more radical opinion that `raw_html` as an alternative to pure ANS content, and wrote this such that you could do it either way. If this is too confusing, let me know and I'll update it to lean more one way than the other. Related to @gengel 's comment too.

robpc:
Changed this to a single array of one element based on feedback from @gengel, @chenghsinw, and others.

scarothers:
I’m really in favor of this. I think it makes a lot of sense to do at the individual content element level and like you said Rob, let downstream platforms decide if they can render the raw_html or not. This will help on AMP pages too, since AMP is really restrictive about HTML/JavaScript and we have a lot of problems with raw_html on it. 

One thing just wanted to raise:
The message if there’s no alternative is in English, which wouldn’t be ideal for our non-English publications. It is probably preferred to have an English language message than no message at all for content that can’t render on a platform, but just wanted to note. (There might be workarounds on the front-end, like if it spots that exact text, render another chunk of text that the client provides in their language instead.)

docs/proposals/2019-02-08 - Fallback Content Elements.md
```diff
@@ -0,0 +1,90 @@
...
+    }
+  ]
+}
+```
+
+# Concerns
+
+## Wouldn't this be better flipped? Where the `raw_html` was available as an alternative to an content_element? 
+
+Yes, honestly, in my opinion that is the "right" way to handle `raw_html`. The `raw_html` element should ONLY be allowed as an alternative to "pure" ANS content elements and in a well-known named one (ex `browser`) to help rendering applications make the right decision on content. The `raw_html` element is an escape hatch to ANS and defeats many of the benefits of ANS to the rendering layer. There should never be `raw_html` at the root level of the content element list. What I am calling the `basic` alternative above should be the content element or collection in an `element_group` and the `raw_html` could be in the alternatives for those applications that can use it.
```
StephanieDClark: This assumes a lot about the ability of adapters to sort through, say, embed code, and build an appropriate set of alternatives.  I mean, adapters would have to know all the things.  

In my opinion, raw_html is a perfectly acceptable root element.  It doesn't always contain "bad" elements, just stuff that it is maybe not worth extracting.

docs/proposals/2019-02-08 - Fallback Content Elements.md
```diff
@@ -0,0 +1,90 @@
+# Alternatives to Raw HTML
+
+# Problem
+
+The content element `raw_html` is often used to inject Javascript into an ANS document which makes it difficult for downstream applications to render natively. Even if the application tries to render it as a embeded browser, there is no guarentee that it will render correctly in all applications. And some applications (ex Apple TV or $50 Kindle) may not have the input or processing capabilities needed to effectively render the experience.
+
+# Proposal
+
+Add a required `alternative` trait to the `raw_html` content element, that contains an array of one content element (allows for multiple later) that can notionally replace the `raw_html` block. The content in the array can only be of type `text`, `image`, `gallery` or `video`. When `upvert`ing this trait should be filled this with a `text` element that has a stock `This content is not available on this platform` message (or similar).
```
StephanieDClark: This trait makes sense, where it can be used.  But it should NOT be required.

chadmoone: Requiring this forces, or at least encourages, content creators to add affordable for non-html clients.

docs/proposals/2019-02-08 - Fallback Content Elements.md
chadmoone: This seems to assume that HTML is always renderable, or that we are dealing with HTML-native clients. This is absolutely not the case. `raw_html` is explicitly placing client-side display logic for one particular client directly into the document. This is obviously accepted in some cases due to lack of better options, but IMHO it shouldn’t be looked at as a valid base level element, at least not without a fallback/default option.

robpc:
> One thing just wanted to raise:
> The message if there’s no alternative is in English, which wouldn’t be ideal for our non-English publications. It is probably preferred to have an English language message than no message at all for content that can’t render on a platform, but just wanted to note. (There might be workarounds on the front-end, like if it spots that exact text, render another chunk of text that the client provides in their language instead.)

I think this is a really good point, but I am not sure what the right approach would be since there is no concept of placeholder for pre-canned messages in ANS that would normally call for support of `i8n` variations. I think that would maybe be worthy of a different discussion (and different change). My statement for the pre-canned message here is only meant to cover upverting documents, but I imagine in the future Ellipsis would handle a similar default message and be aware of language differences in the same way it is for it's own ui. Happy to change the upverting suggestion for a better idea, but I think it is important to include something.

docs/proposals/2019-02-08 - Fallback Content Elements.md
robpc: I agree with Chad that this field being required is important. Without it there is no way to make sure that the issues with `raw_html` are mitigated. If content producers are unhappy with the default content here then we can change default, but in practice they should be using Ellipsis to tailor it for each case, whether that is a message, link, video, image, or whatever. Having the content not show up at all is worse than seeing a stock message in almost every case. If it's not required, then everyone is going to forget to set this field and that defeats the purpose. I don't think it is unreasonable to tell users that if they use `raw_html` it comes at the cost of making sure there is a viable alternative.

gengel: @robpc @chadmoone Are we comfortable saying, then, that any legacy content that has already been saved contains raw_html (of which there is quite a bit) cannot be changed or re-published until an alternative is manually added to each raw_html object? (I am, but it's a tradeoff worth being explicit about.)

What about ingested content from legacy CMSes that store data in HTML (which is almost all of them?) When an automated ingestor pulls this old content in, and can't parse the entire article into ANS elements, what is the correct behavior? 

I suspect the result you'd be incentivizing is that any content where this occurs ends up with a very generic "Missing Element" placeholder image. It's not immediately clear to me that this is better than simply skipping over the raw_html element.

chadmoone: The reason I think it's better is that it's explicit. I think it applies to both newly generated content and imported content.  Right now, what happens in apps is that there is just nothing there at all.

To the end user, I think a clear indicator that there is article content here which can't be rendered on this device is important, because they might be able to try reading it another way. But just as important is internal troubleshooting. I think many of the Arc partners have a more hands-off approach to apps, but this is absolutely a common case on WP side. We get asked all the time which a given chunk of content doesn't show up on the app, or even if we don't hear about it, we notice it ourselves when the article body references something which is nowhere to be found.

Admittedly, this is a crap experience when the bulk of the content elements are actually just `raw_html`.  In this case however, I think I'd recommend:
1. Force at least the default `alternative` value, but just as importantly
2. Set the new `rendering_guides.preferred_method` defined in #198 appropriately, which would cause supporting clients which cannot render `raw_html` to open in a web view (if available).

I hadn't considered this case before, but for the bulk ingestion of purely HTML content which is just going to get dumped primarily/completely into `raw_html` elements (😢), then I think both of the above should be applied.  For legacy content already imported which this applies to, I'm not sure if there is a good way to update the second part retroactively, but maybe someone else has an idea there.

gengel: If part of the recommendation for legacy (or otherwise substantially HTMLish) content is to set `rendering_guides.preferred_method`:`www`, what value does setting the `alternative` add? It seems like once `www` is set, there's no real choice but to open a webview anyway.
  
My worry here is that if by making this field required, it may actually have the opposite effect as intended. The intended effect, AFAICT, is that for each piece of content where someone has decided to manually insert some HTML extract, they should be asked to set a replacement content simultaneously. In this way, non-HTML renderers can insert something in place of the HTML having been safely assured that they are showing what was intended by producer. So far, so good. 

But imagine the human equivalent of the ingestor scenario. A user is asked to (intentionally) insert some HTML in their article because of $BUSINESS_REASONS. Then we inform that they can't publish the article without a replacement, even if there is no reasonable replacement available. So the user does what? My guess is they select "text" mode and input "N/A" or the empty string or whatever it takes to make the error message go away. At that point, the field no longer captures "intended replacement" in a meaningful way, and the apps are back to rendering content with potentially critical stuff missing.

Is it possible that this is less a data model requirement and more a business-logic requirement: that for all *new* content created, either `rendering_guides.preferred_method`:`www` must be set, OR an alternative should be set for each-and-every `raw_html`? This gives legacy ingestors a clean path forward (just set `www` on everything) while also pushing users towards cleaner content.

anscho: The "Classic App" (at the time, just "the iPhone app") used to run off of an RSS feed that didn't support any content type except text, images, and videos. In place of any other embed, it would render "This content is not supported." One of the first requests I worked on was getting rid of those error messages, because it made the app look broken.

I'm much more in favor of transparency than sweeping it under the rug, but I agree that mandating this doesn't scale well for outside and existing content.

What do you guys think about, if no alternatives are provided, replacing the raw HTML with a "view on web" embed? Then the default experience isn't a dead end, and engaged newsrooms can provide a better experience.

robpc: > One of the first Newsroom requests I worked on was getting rid of those error messages, because it made the app look broken.

Silently hiding it has the same effect, or worse they don't notice. I don't think `This content is not supported` is the desired outcome but it makes what is happening clear and hopefully would bug them to use the content properly. The difference with your example is there was no path for the newsroom to correct it. Here if the newsroom is saying that `This content is not supported` is broken then the response is that they didn't properly set the alternative and they should go do that. 

I think these points are valid for not being draconian retroactively, but as @gengel says I am not sure a better way to do it. Without a forced alternative it is easy to fall into the same pattern we have now, where these items fall through the cracks and people are "shocked" to find out they are missing from some platforms. Giving a default message doesn't make it more broken, it just exposes to newsroom users that it is broken which is an improvement. Even better they have the tools to fix it. 

From the statements here it seems like people are imagining a situation where they add the `raw_html` and then an hour after publishing realize that there is a stock message saying the content doesn't work. That's not how this should work. Ellipsis or whatever other tool should be showing this to them as the alternative as they are adding the content. The default message is really just an invitation to put something better. 

Another thought I had over the weekend was having a flag that says the content is not required, which would allow the alternatives to not be provided and signal that it was safe to ignore the content. The upverting could default to this being turned on, but going forward it would be turned off and the default message would be provided. 

I don't love this because of the issues already mentioned, but those could be mitigated by clear messaging in the content editor. This solution would allow for the cases mentioned (sequential `raw_html` content, dynamically imported etc). It also allows the editor to indicate it's not needed, but requires them to make take an action to say it's not needed. There are many ways this could fall through so I am not really advocating for it, but it's the only alternative I can think of.

robpc: > If part of the recommendation for legacy (or otherwise substantially HTMLish) content is to set rendering_guides.preferred_method:www, what value does setting the alternative add? It seems like once www is set, there's no real choice but to open a webview anyway.

@gengel I believe that @chadmoone is saying that you would set a default alternative for most situations, but in situations where that doesn't work (import of almost all html only content, etc) you could set the preferred view as another way to deal with it. These two requests are in our minds different ways of handling the same issue. One is a hammer, the other a scalpel.

robpc: Whoops missed this statement from @gengel

> Is it possible that this is less a data model requirement and more a business-logic requirement: that for all new content created, either rendering_guides.preferred_method:www must be set, OR an alternative should be set for each-and-every raw_html? This gives legacy ingestors a clean path forward (just set www on everything) while also pushing users towards cleaner content.

I think that is what @chadmoone is saying and works for me

gengel: @robpc Cool. In that case, I suggest we drop the `required` attribute for the JSON schema and instead forward that requirement Ellipsis (cc @awong7)

styonsk: I am giving approval based on the decision here to drop the required attribute from the schema. If this decision is changed I'd like to withdraw approval and discuss more.

robpc: Updated the language to reflect @gengel and @chadmoone 's feedback. Should meet @styonsk's concerns, though I would like to point out that this element is still required even if within the spec we will mark it as not to support one very specific exception. 

```
This trait is required unless the `rendering_guides.preferred_method` is set to `website` only.
```

I have concerns about that, but as long as Ellipsis is willing to support the intent there then I am happy to move it forward.

docs/proposals/2019-02-08 - Fallback Content Elements.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Example 2 (future implementation of multiple elements)
+
+The social team is trying to build awareness of the Twitch channel so they are embedding the Twitch stream where normally we would use the Goldfish live video. They provide the normal video as an alternative as well as a note to encourage users to open the twitch channel.
+
+```
+{
+  "type": "raw_html",
+  "content": "<iframe src=\"https://player.twitch.tv/?channel=coolpublishing\" frameborder=\"0\" allowfullscreen=\"true\" scrolling=\"no\" height=\"378\" width=\"620\"></iframe><a href=\"https://www.twitch.tv/coolpublishing?tt_content=text_link&tt_medium=live_embed\" style=\"padding:2px 0px 4px; display:block; width:345px; font-weight:normal; font-size:10px; text-decoration:underline;\">Watch live video from coolpublishing on www.twitch.tv</a>",
+  "alternative": [
```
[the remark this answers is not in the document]
robpc: As you point out, determining if the `raw_html` rendering failed is difficult, but this proposal does not try to address that case. It is intended to be a fallback for rendering platforms that otherwise ignore the `raw_html` element or reject a story because it has that element. The rendering platform should make the determination itself depending on if it supports open-ended html. Many of the rendering platforms other than the website fall into this category (mobile apps, facebook instant, apple news, google news, etc) and `raw_html` can make the article confusing if important content is contained in it and it's not shown. This provides a mechanism to allow for a passible alternative, or to inform the reader if there is something not shown when `raw_html` is not supported.

robpc:
@styonsk @gengel Any issues with this proposal not addressed? Back in March y'all gave feedback and I made what I think were the requested changes, but never got additional feedback or approval. Could you please let me know where you are with this change?

docs/proposals/2019-02-08 - Fallback Content Elements.md
```diff
@@ -0,0 +1,84 @@
...
+}
+```
+
+# Concerns
+
+## Wouldn't this be better flipped? Where the `raw_html` was available as an alternative to an content_element? 
+
+Yes, honestly, in my opinion that is the "right" way to handle `raw_html`. The `raw_html` element should ONLY be allowed as an alternative to "pure" ANS content elements and in a well-known named one (ex `browser`) to help rendering applications make the right decision on content. The `raw_html` element is an escape hatch to ANS and defeats many of the benefits of ANS to the rendering layer. There should never be `raw_html` at the root level of the content element list. What I am calling the `basic` alternative above should be the content element or collection in an `element_group` and the `raw_html` could be in the alternatives for those applications that can use it. 
+
+Realistically though, given momentum and the frequent use of raw_html now, I am not sure that this answer would be tenable, so the proposal is a version of this that is more backwards compatible.
```
styonsk: Does this mean that alternatives will not be required?

gengel:
@robpc My apologies for the delay, Rob. I'd like to see this field be totally optional, regardless of other fields. Otherwise it's effectively a breaking change.

robpc:
@gengel I thought we came to a resolution you were happy with in the earlier discussions. From your comments:

> Is it possible that this is less a data model requirement and more a business-logic requirement: that for all new content created, either rendering_guides.preferred_method:www must be set, OR an alternative should be set for each-and-every raw_html? This gives legacy ingestors a clean path forward (just set www on everything) while also pushing users towards cleaner content.

Are we opposed to making any breaking change in the format? This is a big issue with it as is and I was hoping we would be able to correct it. Even this change you ask for above is a big softening of the initial request. There is certainly value in the proposal without that requirement, but I am concerned that anything not in the spec will not be enforced.

gengel:
Except in cases where it is absolutely necessary, yes, we avoid breaking changes. 

You're right -- you won't be able to rely on incoming data enforcing that alternatives are set. But even if it's optional, having *some* files with this set is better than having *no* files, I think.

robpc:
Agree that something is better than nothing. I will take it out if that is the only way to move forward.

gengel:
Thanks! Sorry again for the slow response.

chadmoone:
Personally, I think making this optional reduces practical usefulness by such a degree that I wonder whether it's even worth implementing at all.  I kind of can't believe I'm saying this, but would it be better to wait until we can implement it the right way?

Though based on history, this would probably mean never?

It's hard to know how to approach improving the spec if it can't evolve in any meaningful way.

robpc:
@gengel Requested changes are made.

PR #200, opened 2019-02-14
EL-2613: Proposal for Dropcaps

michaelguild13:
This is the initial proposal for adding `dropcap`'s to ans content elements of type `text` and `header`

### DIFF
--- a/docs/proposals/2019-02-14_Dropcap.md
+++ b/docs/proposals/2019-02-14_Dropcap.md
@@ -6,11 +6,11 @@ Users want the ability to add a dropcap to paragraphs and headings.
 
 # Proposal
 
-we propose adding a new trait, `formating` to text and heading content elements, which will contain the new property `dropcap`.
+we propose adding a new trait, `formatting` to text and heading content elements, which will contain the new property `dropcap`.
 
 `dropcap` will accept a string value of `letter` or `word`.
- * `letter` implies to apply the dropcap formating to the first letter of the first word of the content element.
- * `word` implies to apply the dropcap formating to the entire first word of the content element.
+ * `letter` implies to apply the dropcap formatting to the first letter of the first word of the content element.
+ * `word` implies to apply the dropcap formatting to the entire first word of the content element.
 
 # Trait in ANS Schema
 
@@ -23,7 +23,7 @@ content_elements: [{
   content: "bla bla bla"
   level: 1
   type: "header"
-  formating: {
+  formatting: {
     dropcap: "letter"
   }
   _id: "KKAE5ETLOVBLXCU7ZI4ELLVUJM"
--- a/docs/proposals/2019-02-14_Dropcap.md
+++ b/docs/proposals/2019-02-14_Dropcap.md
@@ -6,7 +6,7 @@ Users want the ability to add a dropcap to paragraphs and headings.
 
 # Proposal
 
-we propose adding a new trait, `formatting` to text and heading content elements, which will contain the new property `dropcap`.
+we propose adding a new trait, `format` to text and heading content elements, which will contain the new property `dropcap`.
 
 `dropcap` will accept a string value of `letter` or `word`.
  * `letter` implies to apply the dropcap formatting to the first letter of the first word of the content element.
@@ -23,7 +23,7 @@ content_elements: [{
   content: "bla bla bla"
   level: 1
   type: "header"
-  formatting: {
+  format: {
     dropcap: "letter"
   }
   _id: "KKAE5ETLOVBLXCU7ZI4ELLVUJM"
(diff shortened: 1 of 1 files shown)


docs/proposals/2019-02-14_Dropcap.md
```diff
@@ -0,0 +1,44 @@
+# Dropcap for paragraph and heading elements
+
+# Problem
+
+Users want the ability to add a dropcap to paragraphs and headings.
+
+# Proposal
+
+we propose adding a new trait, `formatting` to text and heading content elements, which will contain the new property `dropcap`.
```
chenghsinw: I think `format` is shorter and simpler.

chadmoone:
Yeah, as @anscho mentioned, this is another prime example of something which should be accomplished using the text attribute system. Some of the biggest concerns that I have with this approach are mostly regarding what happens when that would actually be implemented (which I'm probably a little ignorant about).  How challenging does it make upversion when we start adding additional format options outside of the HTML?  And I'm assuming we would then deprecate/remove this `format` property once that happens, because it's redundant/conflicting?

Another option (which I probably need to think through a bit more) might be to approach this by actually adding an initial implementation of that attributes system, but only supporting new formatting options like this for the time being and waiting to a later date (1.0) to move core formatting (bold/italics/links/etc) there and enforce it as the core standard. This would (IMHO) be a much cleaner implementation, and would be much more forward-looking.  I fear we are setting ourselves up for more pain by creating a custom formatting/attributes space which will eventually be replaced by another one.

docs/proposals/2019-02-14_Dropcap.md
```diff
@@ -0,0 +1,44 @@
+# Dropcap for paragraph and heading elements
+
+# Problem
+
+Users want the ability to add a dropcap to paragraphs and headings.
+
+# Proposal
+
+we propose adding a new trait, `format` to text and heading content elements, which will contain the new property `dropcap`.
```
vgritsenko: Did you mean `header`?

https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.9.1/story_elements/header.json

docs/proposals/2019-02-14_Dropcap.md
```diff
@@ -0,0 +1,44 @@
+# Dropcap for paragraph and heading elements
+
+# Problem
+
+Users want the ability to add a dropcap to paragraphs and headings.
```
vgritsenko: What about other textual elements, such as: `blockquote`, `correction`, `interstitial_link`.

michaelguild13: For now, we are just focused on do ing text and header. It's possible those will need to be considered as well.

docs/proposals/2019-02-14_Dropcap.md
michaelguild13: Yes

michaelguild13:
@chadmoone whats the text attribute system? Is this currently being proposed?

chadmoone:
Yes, it's #134.

michaelguild13:
> Yes, it's #134.

Is this getting any momentum?

chadmoone:
I don't know about momentum, but I believe it's supposed to happen once the transition to the new publishing platform is complete. 

Perhaps it could be approached in phases, though, where
* Phase 1: A small subset of styles/attributes (eg `dropcaps`) which do not currently existing in text elements gets implemented using the new system.
* Phase 2: Conversion to the attributes as canonical storage, support for conversion and all formatting through attributes.

What I'm trying to avoid is competing/duplicative approaches to the same problem which will make it harder to untangle in the future.

I don't know what will go into the conversion when we get there, so I'm note sure whether this phased approach presents any issues for process, but it makes sense to me from a consumer perspective.

michaelguild13:
@chadmoone so we should probably put a hold on dropcaps, yeah?
@awong7

docs/proposals/2019-02-14_Dropcap.md
design1online: What about people who want sub or sup scripts? I actually heard someone asking for being able to sup or sup words in the header as well. Would they be able to apply any other additional styles to the dropcap?

So I guess my only concern with this is that if we break this off into it's own **format** section then it might be confusing on what things in the format applies to, especially if we start adding things like color/size/style.

For instance, if we ended up adding format: { style: bold } would that apply to the entire header or just to the dropcap? It might make more sense to put this at the same level that we have the header:

```
content_elements: [{
  additional_properties: {},
  alignment: "left",
  channels: ["basic", "print"],
  content: "bla bla bla",
  level: 1,
  type: "header",
  dropcap: "letter",
  _id: "KKAE5ETLOVBLXCU7ZI4ELLVUJM"
}]
```

PR #201, opened 2019-02-15
EL-2625: Proposal for Table of Contents Anchor links

michaelguild13:
This is the init proposal for TOC and anchor links

### DIFF
--- a/docs/proposals/2019-02-15_TOC_Anchor.md
+++ b/docs/proposals/2019-02-15_TOC_Anchor.md
@@ -2,42 +2,38 @@
 
 # Problem
 
-Users want the ability to have a tabel of contents (TOC) in order to jump to different locations within an article.
+Users want the ability to have a jump links within an article.
 
 # Proposal
 
-We propose adding a new trait, `anchor` to content elements, which will contain the properties of `id` and `label`.
- * `id` will accept a string value which will be used to generate a hash seo friendly id.
- * `label` will accept a string value which will be used as the anchor's display text
+We propose adding two new content types, `Table of Contents` and `Anchor Divider`.
+- Anchor Link List / Table of Contents
+- Anchor Divider
 
-# Trait in ANS Schema
+# Anchor Divider
+A new content type used to create a hidden horizontal rule which will be utilized to auto generate a Link List/TOC. This hidden horizontal rule will have two properties, `text` and `id`.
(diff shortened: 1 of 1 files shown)


michaelguild13:
Has everyone been able to reviews this yet?

michaelguild13:
If no one has issues with this proposal, I would like to merge and start writing the schema.
@gengel @chadmoone

st421:
When I see the name `anchor`, I think of an html tag that can link to anything (i.e. it would not be clear to me that this is limited to items within the current document). I think something like `internal_link` might make the functionality more clear. On the other hand, it's not uncommon for clients to want elements that do have links to external content so something more generic that would allow for internal or external linking might be useful.

chadmoone:
I previously approved, and still good with this. Personally I think the term `anchor` is fine because it's part of ANS, and it makes sense to me that it would be limited to anchor to an ANS content element.

gengel:
@design1online An interstitial link is an element that represents a link out to another story. 

An anchor *property* on an element represents a way of linking *to* an element within a story. It's roughly equivalent to an HTML name/id. (You could populate the name/id in your example from the anchor property on an element.) Because it's a property, it can be applied to any content element, including an interstitial_link.

design1online:
@gengel true, but we also have clients that are using it to reference external links (RTL is one that actually wants functionality for their interstitial links to open in an external window, and I believe Raycom has asked for this before too).

For instance, in Ellipsis you can create a link to any URL and that is saved as an interstitial link in the ANS. I'm not sure why we need another ANS element for what is essentially already doing the same thing. I think updating the interstitial link to include the anchor/target/title/id/name properties makes more sense.

gengel:
@design1online 

I think the use of the name "anchor" here and the fact that HTML uses the "a" tag for two different things--to mark text as a link-to-something-else, and as something-that-can-be-linked-to --is causing some confusion. 

This is not creating a new way to link out to another story -- it's creating a reference point for other links to link to, and a mechanism for editors (and if they want, front-end renders) to generate a table of contents for a document. 

It's also proposed to be implemented as a trait on any content element. So you could assign an anchor to an interstitial link in Ellipsis, and this would effectively give you what you're suggesting (a name/id/title property on an interstitial link.)

michaelguild13:
@gengel @chadmoone @design1online
Proposal has been updated

PR #203, opened 2019-03-18
Links list proposal

gengel:
A proposal for embedded and curated "See More" boxes

### DIFF
--- a/docs/proposals/2019-03-18-Links-List.md
+++ b/docs/proposals/2019-03-18-Links-List.md
@@ -2,21 +2,23 @@
 
 Organizations have requested a way to include sidebars, related content boxes, or other recirculation links within the body of an article. A curated solution -- where users have reasonable control over the presentation and selection of the items on an individual -- is highly-desirable.
 
-Existing solutions are either too generic (`element_group`, `story`) to interpret well in a UX, or too limited (`related_content`, `promo_items`) to be useful in this capacity, or cannot be meaningfully extended in a non-breaking way.
+Existing solutions are either too generic (`element_group`, `story`) to interpret well in a UI app, or too limited (`related_content`, `promo_items`) to be useful in this capacity, or cannot be meaningfully extended in a non-breaking way.
 
 
 ## Proposal
+
+Extend the `interstitial_link` element to support more robust presentation, adding an optional `description` and `image` elements. The image should be considered part of the link. The existing `content` trait will continue to be considered the "main text" of the link, and continue to be required. (E.g. a user interacting with the title or image would expect to navigate to the destination of the link.) The `description` field will be an ANS `text` element and may contain limited formatted text (using inline HTML only.) The description field may or may not be considered part of the link itself, depending on implementation.
 
 Add a new content element `link_list` that contains a narrowly-targeted set of functionality around inline recirculation boxes. Applicable to stories only.
 
 A links_list may have:
 * A plaintext title
 * A subtype (selected from an enumerated-list configured on a per-organization basis)
-* A sequence of up to 10 and at least one `link_list_item`s, each of which has:
+* A sequence of up to 10 and at least one `interstitial_link`, each of which has:
     * A subtype
-    * A plaintext title (required)
-    * A url (required)
-    * A description (ANS `text` element)
+    * A plaintext title (required) (`content`)
+    * A link (required) (`url`)
+    * A description (`description` -- ANS text element)
     * An image (ANS `image` or `reference` to an image)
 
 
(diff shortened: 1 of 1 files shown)


gengel:
Temporarily rescinded -- this needs more changes

gengel:
Reopened!

docs/proposals/2019-03-18-Links-List.md
```diff
@@ -0,0 +1,112 @@
...
+      "subtype": "sidebar",
+      "title": "More 'Biting' Stories",
+
+      "items": [
+        {
+          "type": "links_list_item",
+          "subtype": "blue-label",
+
+          "title": "Lorem ipsum",
+          "link": "http://www.google.com",
```
chenghsinw: From the last email for Prisa, looks like they need a way to configure additional data for the link: rel="nofollow" and target="_blank". Do u think we should have additional_propertties in links_list_item for custom requirements or introduce specific properties for the needs?

gengel: I think they should key off of subtype if they really need it.

yury-sannikov: It seems it really depends on how many parameters users want to control. For instance, for RED/BLUE/BLACK label color and nofollow/blank settings, we will have to provide 12 subtypes.

chadmoone: Are you suggesting this be a text item which can have an `<a>` tag? I think that opens this up far more than we want to.

chenghsinw: Do u mean they have to define subtypes something like these:
- blue-label-nofollow
- blue-label
- red-label-nofollow
- red-label
- green-label-nofollow
- green-label
- ...
It looks like more complex and each org might implement the pattern differently which will make Ellipsis hard to standardize the UI.

gengel: @chadmoone No tags. The entire entry is a link to the URL specified.

gengel: @chenghsinw @yury-sannikov We should not assume a specific rendering. We're trying to capture a general version of the use case. The front-end can key off of the subtype to determine labels, colors, nofollow, whatever other presentation options they want.  If there is a proven widespread demand for labels or nofollow, specifically, we can add those options at a later time.

StephanieDClark: @gengel This looks alot like "interstitial_link".  Would this be intended as a replacement, with a new subtype of "interstitial_link"?

gengel: It's a similar use case, but not quite the same. The intention is that this is it's own, new element type for groups of links with a title for the box/section, a plaintext title for each link, optional images, but no embedded HTML or formatting. 

This is a little bit more structured, and more of an "aside" to the main content.

scarothers:
Agree about link_list as the name. Overall makes sense to me – I've seen a lot of clients want to curate related stories within the body of the article (specifically able to place them at a certain point in the body of the article, which you can't do with related_content), and I think this makes sense for that use case. Two questions: 

* With title being plaintext, does that mean it couldn't include italics? I think people might need that if the headline has an italicized word in it ... should it just be the same as whatever headlines.basic accepts? 
* I agree/understand about this not needing to be a story reference (because it makes the payload so big). Sometimes clients do want to display data about the related story though – like section or date. Should that just be considered something that's not supported by *links_list*?

docs/proposals/2019-03-18-Links-List.md
gengel: We did call the link "url" over in interstitial, though, so I should probably change the field here to be called the same.

gengel:
@scarothers As described here, this proposal would not cover italics or additional metadata. 

Italics is probably desirable, but it's hard to capture safely without supporting all of #134

scarothers:
@gengel makes sense. if clients usually use italics in headline style, we can suggest they use single quote marks instead here.

docs/proposals/2019-03-18-Links-List.md
```diff
@@ -0,0 +1,113 @@
...
+  "headlines": {
+    "basic": "Man Bites Dog"
+  },
+
+  "content_elements": [
+    {
+      "type": "link_list",
+
+      "subtype": "sidebar",
+      "title": "More 'Biting' Stories",
```
StephanieDClark: I don't think the title for the type or the items should be required.  

Use cases I see for this include: 
- support the use of a "read more" list. 
In this case, the set of links has a title, "Read More:", but the individual links don't.  See https://www.washingtonpost.com/opinions/2019/03/25/trump-won-with-illicit-help-he-abused-his-power-his-ag-is-blocking-full-reckoning/?utm_term=.cfcdb102bdb2  

- replace the use of content element type "interstitial_link".  
In this case, in interstitial link is really a set of 1, with no headlines.

gengel: What would the link text be if not the title?

robpc: I think maybe what @StephanieDClark is getting at here is there are examples where there is supporting text for a link that is not the link itself? I know that historically this has been available (especially with interstitial links) but was not able to find a recent example.

RyanGladstone: From the journalist's perspective, the interstitial_link content element requires fewer steps than using a links list with a _interstitial_link_ subtype and no headline

st421:
I find this approach confusing. As I mentioned over [here](https://github.com/washingtonpost/ans-schema/pull/201#issuecomment-476356704), people want to add links to single items and I think there should be a way to do so without putting the items in a `link_list`. If elements had a field where you could not only add a link but also specify if the link was to (1) a place internal to the current document (2) another ANS document (so a canonical URL can be used) (3) anywhere else, then it seems like an element group would work when you need a group of linked items, but you'd also be able to distribute your linked items throughout a story if you wanted to (okay, the `element_group` type lacks a `title` but probably it should have one either way).

gengel:
Per feedback from @st421 and @StephanieDClark, I'm going to revisit this to make the relationship between `link_list` and `interstitial_link` more explicit

gengel:
@st421 @StephanieDClark @styonsk @jmjakes @yury-sannikov 

Re-opened. Now uses `interstitial_link` and extends that element.

PR #204, opened 2019-03-19
Implementation of Divider

gengel:
Implementation of divider element schema, which was proposed and accepted in the past but never implemented. (See https://github.com/washingtonpost/ans-schema/blob/master/docs/proposals/2018-01-19_divider_element.md )

Also moves next release version of ANS to 0.10.0.


chenghsinw:
Just curious why we go up to 0.10.0 directly not 0.9.2?

gengel:
@chenghsinw Because this version will have so many new features!

PR #207, opened 2019-04-05
Custom embed schema implementation

gengel:
Implementation of #205


src/main/resources/schema/ans/0.10.0/story_elements/custom_embed.json
```diff
@@ -0,0 +1,68 @@
...
+        },
+
+        "config": {
+          "title": "Embed Configuration",
+          "description": "Arbitrary configuration data generated by a plugin. Users are responsible for maintaining schema.",
+          "type": "object",
+
+          "additionalProperties": false,
+          "patternProperties": {
+            "^(?!(referent|type)$)[a-zA-Z0-9_]*$": {}
```
yury-sannikov: what about `version` field or it's okay to have it?

gengel: @yury-sannikov Good catch; just updated it

src/main/resources/schema/ans/0.10.0/story_elements/custom_embed.json
```diff
@@ -0,0 +1,68 @@
...
+          "type": "string",
+          "maxLength": 128,
+          "minLength": 1
+        },
+
+        "url": {
+          "title": "Embed Provider URL",
+          "description": "Provider URL for this embed item. When concatenated with Embed ID, should produce a URL that returns json metadata about the embedded content.",
+          "type": "string",
+          "maxLength": 128,
```
chenghsinw: I feel 128 length might not be long enough to cover most of the cases:
https://111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
Should we change it to 255?

gengel: Good catch! Looks like I copy+pasted id without thinking. I've bumped to 512

PR #208, opened 2019-04-23
Audio and Audio Stream proposal

williamslm:


docs/proposals/2019-04-23 - Audio.md
```diff
@@ -0,0 +1,50 @@
...
+Add embed_html, series, and audio_type to audio object.
+
+
+## Embed_html
+
+HTML for the embedded audio player
+
+## Series
+
+Unique ID for podcast parent objects.  A series is a collection of related podcasts.
```
design1online: When you say series it makes me think of an audio player that could potentially play multiple podcasts (ie show a tracks and you select which one you would like) but this configuration have an embed html for each one of them. Is that your intention? Or is the idea of a series just to provide a way to query for related audios?

docs/proposals/2019-04-23 - Audio.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Audio_type
+
+An enumeration of types of audio which will include a "podcast" as one of the types
```
st421: What will the other types be? I feel like other types might be less well-defined, in which case a new field isn't necessarily justified (could use `subtype`).

docs/proposals/2019-04-23 - Audio.md
```diff
@@ -0,0 +1,50 @@
...
+# Proposal
+
+Add embed_html, series, and audio_type to audio object.
+
+
+## Embed_html
+
+HTML for the embedded audio player
+
+## Series
```
st421: How might this be used for non-podcast types? Does this name make sense for other types of audio?

docs/proposals/2019-04-23 - Audio-Stream.md
```diff
@@ -0,0 +1,50 @@
...
+# Proposal
+
+Add duration and ad enabled boolean to each stream element inside of the streams array.
+
+
+## Duration
+
+Runtime of audio stream in milliseconds
+
+## Ad_enabled
```
st421: `ad_enabled` says to me that one _could_ put an ad in the stream, not that an ad is already in the stream. `contains_ad` or something like that seems clearer.

docs/proposals/2019-04-23 - Audio.md
design1online: Yeah I was thinking something like `category` might make more sense for that same reason ^

PR #209, opened 2019-05-07
Extends custom embed

gengel:
Allows `custom_embed` element in `promo_items` and `related_content` traits.


src/main/resources/schema/ans/0.10.1/traits/trait_promo_items.json
```diff
@@ -15,13 +15,19 @@
         },
         {
           "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.1/story_elements/raw_html.json"
+        },
+        {
+          "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.1/story_elements/custom_embed.json"
         }
       ]
+    },
+    "embed": {
```
yury-sannikov: Do we need this? embed will be covered by catch-all `".*"` below

gengel: Ah geeze this is a mistake

PR #210, opened 2019-05-09
EL-2663: Schema for format dropcaps

michaelguild13:
Format, dropcap schema based on PR
https://github.com/washingtonpost/ans-schema/pull/200

### DIFF
--- a/src/main/resources/schema/ans/0.10.0/story_elements/header.json
+++ b/src/main/resources/schema/ans/0.10.0/story_elements/header.json
@@ -21,7 +21,9 @@
     "alignment": {
       "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.0/traits/trait_alignment.json"
     },
-
+    "format": {
+      "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.0/traits/trait_format.json"
+    },
 
     "additional_properties": {
       "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.0/traits/trait_additional_properties.json"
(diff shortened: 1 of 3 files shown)


gengel:
I think this needs to be against at least 0.10.1 now, better even 0.10.2

Also, tests need to be added or altered

michaelguild13:
> I think this needs to be against at least 0.10.1 now, better even 0.10.2
> 
> Also, tests need to be added or altered

Ok, I'll update to 0.10.2 and check tests

src/main/resources/schema/ans/0.10.0/story_elements/header.json
```diff
@@ -21,7 +21,9 @@
     "alignment": {
       "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.0/traits/trait_alignment.json"
     },
-
+    "format": {
```
yury-sannikov: I think header should not have this element

src/main/resources/schema/ans/0.10.0/story_elements/text.json
```diff
@@ -21,6 +21,9 @@
     "alignment": {
       "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.0/traits/trait_alignment.json"
     },
+    "format": {
```
yury-sannikov: @gengel do you think it make sense to call it format? Maybe something like `block_properties`. Format kind of confusing with text format.

yury-sannikov: Also, file name traits/trait_format.json sounds cryptic to me

src/main/resources/schema/ans/0.10.0/story_elements/header.json
michaelguild13: I was originally discussed to have it but I am fine with removing that as well

src/main/resources/schema/ans/0.10.0/story_elements/text.json
michaelguild13: Agreed, just following everyone's file naming convention

michaelguild13: `block_properties` works...or `formatting`? I am open. This probably should've been more covered during proposal

michaelguild13:
Opened a new PR that targets 0.10.2 with test story
https://github.com/washingtonpost/ans-schema/pull/212

PR #212, opened 2019-05-10
EL-2663: Schema for format dropcaps

michaelguild13:


michaelguild13:
Yes, it's unrelated to my changes. It has to do with the custom_embed

src/main/resources/schema/ans/0.10.2/traits/trait_block_properties.json
```diff
@@ -0,0 +1,15 @@
...
+  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.2/traits/trait_block_properties.json",
+  "title": "Block properties",
+  "description": "Block properties for style formatting content elements",
+  "type": "object",
+  "properties": {
+    "dropcap": {
+      "title": "Dropcap",
+      "description": "Style the first letter or first word with a dropcap",
+      "type": "string",
+      "enum": ["letter", "word"]
```
yury-sannikov: I think per @awong7 we're not going to support `word` type

yury-sannikov:
overall, this PR has not enough tests. It should have both positive and negative tests for each use case

michaelguild13:
> overall, this PR has not enough tests. It should have both positive and negative tests for each use case

Sure thing. Wasn't fully aware of what tests were needed for this. I'll update accordingly.

src/main/resources/schema/ans/0.10.2/traits/trait_block_properties.json
michaelguild13: @yury-sannikov in that case, should dropcap be a bool?

michaelguild13:
@yury-sannikov I've updated the test to include bad cases as well

PR #213, opened 2019-05-13
CA-1512: tags subtype proposal

wapo-omar-khan:


docs/proposals/2019-05-13_Tags_Subtype.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Example
+```json
+{
+  "_id": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
+  "type": "story",
+  "version": "0.10.1",
+
+  "taxonomy": {
+    "tags": {
```
anscho: Should this be `tag` or an array?

jpblair:
Does this change anything other than adding a `subtype` field to the existing tag objects?

gengel:
@jpblair No

docs/proposals/2019-05-13_Tags_Subtype.md
wapo-omar-khan: according to the schema, it's an object called tags

wapo-omar-khan: and i just remembered that elasticsearch mappings consider elements and arrays of that element the same type. arrays is right

PR #217, opened 2019-07-12
Focal Point Proposal

jimas14:
This PR includes the proposal for adding focal point to ANS for use in images.

### DIFF
--- a/docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
+++ b/docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
@@ -74,7 +74,10 @@ The above example sets a focal point at the intersetion of 460px from the LEFT e
 }
 ```
 
-### Example gallery ANS overriding the included image's focal point using additional_properties of the image
+## Gallery Focal Point Per Image Over-ride
+A Gallery can provide an individual over-ride per image in the gallery. These will be represented as `gallery_properties` on the Image ANS root. The new `gallery_properties` property is only valid in Image ANS within the context of a gallery.
+
+### Example Gallery ANS over-riding the included image's focal point using `gallery_properties` of the image
 ```
 
 {
@@ -137,10 +140,6 @@ The above example sets a focal point at the intersetion of 460px from the LEFT e
 	"content_elements": [{
 		"_id": "2K2QZFZXT5BPBPGFPKJIGL7D5Q",
 		"additional_properties": {
-      "focalPointOverride": {
-        "x": 700,
-        "y": 600
-      },
 			"fullSizeResizeUrl": "/photo/resize/Oyx05jwTKr3nakiLHktQnO2eD3w=/arc-anglerfish-local-localhost/public/2K2QZFZXT5BPBPGFPKJIGL7D5Q.jpeg",
 			"galleries": [{
 				"headlines": {
@@ -167,10 +166,16 @@ The above example sets a focal point at the intersetion of 460px from the LEFT e
 		"credits": {
 			"by": []
 		},
-    "focal_point": {
-      "x": 1146,
-      "y": 806
-    },
+		"focal_point": {
+			"x": 1146,
+			"y": 806
+		},
+		gallery_properties: {
+			"focal_point": {
+				"x": 700,
+				"y": 600
+			}
+		},
 		"geo": {},
 		"height": 4302,
 		"image_type": "illustration",
--- a/docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
+++ b/docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
@@ -28,6 +28,15 @@ The X coordinate's range is 0->Image Height. The Y coordinate's range is 0->Imag
 }
 ```
 
+### Example Focal Point
+```
+"focal_point": {
+  "x": 460,
+  "y": 300
+}
+```
+The above example sets a focal point at the intersetion of 460px from the LEFT edge of the image, and 300px from the TOP of the image.
+
 ### Example image ANS with focal point
 ```
 {
--- a/docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
+++ b/docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
@@ -12,7 +12,7 @@ Editorial users wish to be able to declare a focal point for images. Additionall
 ## Proposal
 We propose adding a focal point trait for use in images. The focal point is a single pixel position identified on an image will be stored by a set of two integer values: X and Y coordinates. The focal point field will appear in the top-level of image ANS.
 
-Parent documents can override an image's focal point. Gallery overrides will appear in the image's additional properties (as caption/headline overrides are). Story and thumbnail overrides will be done through referent_properties on the image.
+Parent documents can override an image's focal point. Gallery overrides will appear in the image's new `gallery_properties` property, which is only valid in the context of an Image ANS within a gallery. Story and thumbnail overrides will be done through referent_properties on the image.
 
 ### ANS Structure
 Focal point values are X, Y coordinates that represent the visual 'center' of the image. When cropping/resizing images, renderers will use the focal point to find the center of the image.
--- a/docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
+++ b/docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
@@ -10,12 +10,16 @@ Editorial users wish to be able to declare a focal point for images. Additionall
 - Goldfish users can set a focal point on the video cover image
 
 ## Proposal
-We propose adding a focal point trait for use in images. The trait will be represented by a set of two integer values: X and Y coordinates. The focal point field will appear in the top-level of image ANS.
+We propose adding a focal point trait for use in images. The focal point is a single pixel position identified on an image will be stored by a set of two integer values: X and Y coordinates. The focal point field will appear in the top-level of image ANS.
 
 Parent documents can override an image's focal point. Gallery overrides will appear in the image's additional properties (as caption/headline overrides are). Story and thumbnail overrides will be done through referent_properties on the image.
 
 ### ANS Structure
-Focal point values are X, Y coordinates that represent the visual 'center' of the image. When cropping/resizing images, renderers will use the focal point to find the center of the image. The X coordinate's range is 0->width. The Y coordinate's range is 0->height.
+Focal point values are X, Y coordinates that represent the visual 'center' of the image. When cropping/resizing images, renderers will use the focal point to find the center of the image.
+
+The X-Y coordinates refers to the X and Y axis that are perpendicular lines intersecting at point referred to as the "focal point". The X axis is horizontal line and the Y axis the vertical line, and represent the LEFT, TOP of the image.
+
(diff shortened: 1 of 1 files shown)


chadmoone:
A few questions: 
1. Why not use a rectangle? Realistically, images often have several vital focal points, which cropping should not violate. Example, a photo of three people. You could set the focal point on the center person, but that gives the rendering layer no information about the other two faces which should not be cropped. A rectangle is a much better approach to this, and is used in other systems to achieve this. 
2. Where is 0,0? Maybe I missed but did not see this defined. Must be clearly documented. 
3. Why not use percentages instead of pixel values tied to the original dimensions of the asset? I can imagine no shortage of situations where a resize could occur, and it seems extremely easy for this metadata to fall out of sync. Percentage would always be valid unless the image gets cropped along the way, in which case all bets are off regardless. Additionally, I think it would be easier to reason about in the client. When dealing with images, we’re typically sending them through a resizer, and, even beyond the above, it seems more error prone to require math to be based on the original dimensions when we’re now working with an image asset of new dimensions.

epascarello:
Did you look at what Ellipsis does? It is an object with min and max. It was built to be a point, but it was coded to be a rectangular selection.

jimas14:
The proposal was written by Michael Holm who is now on vacation. I'll review this today and discuss with @awong7 about rectangularization. Given how thumbor uses focal points, focal point this proposal should have been a rectangle.

@epascarello Thumbor's focal point filter takes two sets of coordinates: top-left & bottom-right of a rectangle. Is that what the min and max values represent here?

@chadmoone To answer #2, the origin is top-left. I'll add that to the docs. #3 is also a good point that I'll consider. AFAIK the focal point dimensions were meant to be applied the photo's original dimensions. Thumbor's example here https://thumbor.readthedocs.io/en/latest/focal.html gives a set of top-left/bottom-right coordinates based off the original resolution, and then crops down to 400x100 with that focal point in mind.

awong7:
Adding some extra detail:
**1) Focal Point is different from cropping**:
- Focal point has a different use case from cropping. We recommend focal point for more general use to direct the resizers to resize around this point. However crops are for when the editorial focus of the image changes. 
- For example, when Melania Trump went to the Texas flood zone in stilettos, the full-length picture of her was cropped to her feet because the true content of that image was her feet, not her whole body. The image can theoretically end up in so many different presentations that it would be too risky to use just a focal point and hope that the image will always truly present the image to the viewer that emphasized her shoes over other parts of the image. In those cases, when the editorial focus of the image is so specific and the meaning of the image has changed, that's when crop is appropriate. Otherwise focal points are appropriate to set the general place where the image should be   resized around.

**2) Current focal point implementation uses an x, y coordinate to construct the rectangle the resizer provides**
- Ellipsis is the only app that currently lets content producers set the focal point. In Ellipsis, content producers set an x,y coordinate as the focal point. 
- The spot on the image will be used by the automatic resizer as the center point to resize the image around. 
- Once the point is identified and returned as part of the photo record, the front end folks build a rectangle. They can use the following three inputs to calculate the rectangle: (1) that focal point (x,y) location; (2) the size of the full size image in width and height; and (3) the desired size of the image to calculate the rectangle. Below is an example of how this works:

> That image is 6923 pixels wide and 3577 pixels high. Let’s say we want a square 300x300 close-up of the temple, and the focal point is identified by the user as 1150x500. So the upper left of the rectangle we want is 1050x400, and the lower right is 12500x600.
> 
> Then the instructions to the resizer are as follows:
> 
> https://thewashingtonpost.com/resizer/Dcw3j_KcDaxzhb-1GJXPfY4Mf78=/1050x400:1250x600/lawyerdrummer.com/wp-content/uploads/2016/07/zep-6.jpg

- Here are Arc Learning Center docs around focal point: https://redirector.arcpublishing.com/alc/arc-products/ellipsis/user-docs/ellipsis-using-focal-points-set-in-stories-with-the-resizer/. (You will need to be logged into Arc apps in order to access it, or DM me.)

**3) Why we're adding focal points to ANS**
- The focal points in Ellipsis are living in additional properties. We should move them to an appropriate place in ANS.
- Anglerfish is adding the ability to add focal points for gallery cover images. I imagine the same feature will be added to Goldfish for video thumbnails.
- Customers have requested a way to set a global focal point a.k.a. set the focal point once on an image and have that focal point flow into the authoring apps when the image is set. That way when a photo editor adds an image to Anglerfish that, say, will be used in multiple pieces of content for, say, breaking news coverage, the photo editor will set the global focal point on the image. Then when that image is added as the promo image in a story and the cover image in a gallery, the content producers working in the story and gallery won't need to set a focal point in Ellipsis and Anglerfish. The global focal point will be the story and gallery focal points. A content producer could also override the global focal point in Ellipsis or on the Anglerfish gallery edit page to set the focal point to different values.

gengel:
After some discussion on the AF and product side, we've determined that the broader feature needs more design input before it can move forward. The outcome of that will likely impact this proposal, so let's pin this until the outcome of that conversation.

gengel:
@chadmoone 
1) We determined on the product side that rectangle was not desirable, so we're moving ahead with a point.

2) @jimas14 We should definitely be explicit about this

3) No opinion, I think either is fine

awong7:
To further explain Greg's response to #1, product, design, and newsroom integration (Aaron) decided that...

- Editorial users setting a point rather than a rectangle was more intuitive to users. Rectangles are too close to cropping, so would be confusing to users to understand the difference between focal point and cropping. 
- Although they are setting a point, the point is used among other factors (the size of the full size image in width and height and the desired size of the image) to calculate the rectangles needed for the front-end.

gengel:
Let's take this convo to Slack

docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
```diff
@@ -0,0 +1,255 @@
...
+## Usage Patterns
+- Anglerfish users will set a global focal point for an image
+- Anglerfish users can set a focal point over-ride for a gallery cover image
+- Ellipsis users can set a focal point on any image, be it featured image, or an image in the body
+- Goldfish users can set a focal point on the video cover image
+
+## Proposal
+We propose adding a focal point trait for use in images. The focal point is a single pixel position identified on an image will be stored by a set of two integer values: X and Y coordinates. The focal point field will appear in the top-level of image ANS.
+
+Parent documents can override an image's focal point. Gallery overrides will appear in the image's additional properties (as caption/headline overrides are). Story and thumbnail overrides will be done through referent_properties on the image.
```
chenghsinw: Any reasons why we don't wanna use the same way (referent_properties) for Gallery?

michaelholm: I have proposed this and @gengel made an argument that it will be difficult to implement.

scarothers: I'd also be curious why ... it seems like with this implementation, downstream consumers (e.g. front-end websites) will have to know that they need to treat focal point data differently when it comes from a gallery vs. when it comes from anything else.

docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
```diff
@@ -0,0 +1,255 @@
...
+  "additional_properties": {
+    "has_published_copy": true,
+    "is_published": true,
+    "publish_date": "2019-07-11T21:52:06.790Z"
+  }
+}
+```
+
+## Concerns
+No concerns at this time.
```
scarothers: Has anyone tried using this proposed ANS data format with a resizer (e.g. Thumbor) to actually create images with a different size/aspect ratio that use this focal point? That would be my only request/concern – make sure it's reasonably compatible with the main use cases. 

For example, let's say there's an Anglerfish image that's 3000x2000 px wide (note this is a 3x2 ratio). The focal point is set at 1000x1000. The website needs to render a version of this image in a 1x1 ratio, centering around the focal point. It needs to have a few different sizes of the image in order to serve the right size to desktop, tablet, and mobile users (let's say 600x600, 300x300, and 100x100 px images, respectively). 

Is the data we're proposing including in the ANS here (the focal point as an X, Y coordinate in pixels on the original image) enough information for a front-end developer to create different crops of the original image – centering on the focal point – with different aspect ratios and pixel sizes? 

If yes, then I think this is good. I would just like someone to confirm we think this use case can be handled adequately with this data.

docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
gengel: @chenghsinw In an ideal world, yes, galleries would store `reference`s to images and put their overrides in `referent_properties`

However, for better or for worse, Anglerfish galleries use `image` objects with a subset of `image` properties in them. These image objects are stored as-is in Content API and not denormalized (since, in a way, they already are.) 

For downstream consumers inspecting a story or a video, focal point overrides are completely transparent. The consumer simply checks `.focal_point` and the correct value is in place based on the context.  However, with galleries, the consumer still has to check in two places for the focal point value -- `focal_point` and `referent_properties.focal_point`.  So changing the field name, IMHO, doesn't bring more clarity or consistency. On top of that, the Anglerfish paradigm to-date for field overrides has been to put them in fieldnameOverride. Changing that means there's now an inconsistency between old fields in AF and new fields in AF, on top of the inconsistency between AF and other apps.

@chenghsinw If you see a way around this, I am open to suggestions

chenghsinw: @gengel Since u mentioned the `fieldnameOverride`, how does it work today? I don’t see it’s in the ans schema.

chenghsinw: nvm, I think u're referring the overriding using the additional_properties

chenghsinw: I can see the challenge here, but personally I think this kind of rules should be explicitly defined in the schema, so this not the knowledge only a small portion of ppl know. Base on my experience with the PS side, when they work on data ingestion, this ANS schema is the info they rely on heavily. The different behavior for handling field overriding can be overlooked easily.

gengel: 100% agree. I didn't even realize this was the behavior until I started digging into this proposal

docs/proposals/2019-07-12_Image_Focal_Point_Proposal.md
design1online: I second these concerns. When I look at the docs for the resizer docs in the ALC it looks like the front end will be expected to calculate and return the new dimensions for the image but it's not clear how this is supposed to be determined. Should the focal point always be 0,0 in the center or on a specific corner of the photo. https://rtl.arcpublishing.com/alc/arc-products/ellipsis/user-docs/ellipsis-using-focal-points-set-in-stories-with-the-resizer/

chadmoone: This is sort of what I was getting at earlier, but my feeling based on the response was that this should be, by design, explicitly unknown client-side rendering details and outside of the purview of ANS. Different clients will have different rendering conditions and may use different resizes, and will make a call on their end.

gengel:
Thanks for the feedback, everyone! I know it's a slog to get through this process, but I do strongly feel that the finished design is better as a result.

PR #218, opened 2019-07-31
Add focal point trait and gallery_properties trait

michaelholm:
First crack at it. Let me know your thoughts.


src/main/resources/schema/ans/0.10.3/traits/trait_focal_point.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.3/traits/trait_focal_point.json",
+  "title": "Focal Point",
+  "description": "Coordinates that represent the visual 'center' of the image",
+  "type": "object",
+  "$ref": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.3/utils/dictionary.json"
```
gengel: This should not be a plain dictionary -- you should specify explicit x and y fields here instead

src/main/resources/schema/ans/0.10.3/traits/trait_gallery_properties.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.3/traits/trait_gallery_properties.json",
+  "title": "Has gallery properties",
+  "description": "An object for over-rides for images when images are used in a gallery. Example over-ride is trait_focal_point.",
```
gengel: overrides

gengel: Probably could use a longer description explaining how the overrides should be used

src/main/resources/schema/ans/0.10.3/utils/focal_point.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.3/utils/focal_point.json",
+  "description": "Coordinates representing the 'visual center' of an image. The X axis is horizontal line and the Y axis the vertical line, with 0,0 being the LEFT, TOP of the image.",
+  "type": "object",
+  "properties": {
```
gengel: I would just move these fields into the trait_focal_point file, I don't think there's a need for a separate file

tests/fixtures/schema/0.10.3/image-fixture-good.json
```diff
@@ -47,6 +47,10 @@
     ]
   },
   "image_type": "photograph",
+  "focal_point": {
+    "x": 460,
+    "y": 220
+  },
   "additional_properties": {
```
gengel: Can you add 1 or 2 failing focal point cases? E.g. x or y with a string, or a field name other than x or y

And invoke the test here: https://github.com/washingtonpost/ans-schema/blob/master/tests/schema-tests-010.js#L820

PR #220, opened 2019-09-30
Content by reference in Link Lists Proposal

taiweituan:
COMP-3310 - ANS proposal: content by reference in Link Lists
https://arcpublishing.atlassian.net/browse/COMP-3310

This PR includes a proposal for adding `story`, `video`, and `gallery` to interstitial link. An implementation PR will follow upon approval of this proposal.

### DIFF
--- a/docs/proposals/2019-09-30 - Content by reference in Link Lists.md
+++ b/docs/proposals/2019-09-30 - Content by reference in Link Lists.md
@@ -6,17 +6,41 @@ Currently, editors can only add URLs in `interstitial_link` to Link Lists. This 
 ## Proposal
 We propose to add additional traits, `Story`, `Video`, and `Gallery`, to `interstitial_link`, so we can add more types of elements by reference to link lists in addition of `image`.
 
+However, unlink `image`, which is used only for presentation purpose, the `story`, `video`, and `gallery` will be used for a link to it's own content. Therefore, the `url` field should become unavailable when either of the reference to `story`, `video`, and `gallery` are being used.
 ## Example
 ```JSON
 {
   "type": "interstitial_link",
   "subtype": "",
   "content": "British boxer bites opponent",
-  "url": "https://www.washingtonpost.com/sports/2019/04/01/british-boxer-suffers-dental-breakdown-bites-opponent-ring/",
   "description": {
     "type": "text",
     "content": "A <em>British boxer</em> has had his license suspended after being disqualified for biting during a <strong>heavyweight</strong> fight."
   },
+  oneOf: [{
+    "url": "https://www.washingtonpost.com/sports/2019/04/01/british-boxer-suffers-dental-breakdown-bites-opponent-ring/",
+    "story": {
+      "type": "reference",
+      "referent": {
+        "type": "story",
+        "id": "BBB"
+      }
+    },
+    "video": {
+      "type": "reference",
+      "reference": {
+        "type": "video",
+        "id": "CCC"
+      }
+    },
+    "gallery": {
+      "type": "reference",
+      "reference": {
+        "type": "gallery",
+        "id": "DDD"
+      }
+    },
+  }],
   "image": {
     "type": "reference",
     "referent": {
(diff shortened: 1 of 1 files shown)


gengel:
Can you explain what an intended rendering of the example Link List you provided might look like?

taiweituan:
c2de2419-0c52-44ba-8d7b-a44e8dec213a
This is the current design with `Link Image` only. Presumably we will add additional fields for story, video, and gallery.

chenghsinw:
Base on my understanding, the purpose to extend the interstitial_link to allow using a story or other reference is for having an alternative way of using the "url." This proposal seems to mix it up with the usage of "image." Could u confirm?

taiweituan:
Confirming with Ryan

taiweituan:
Confirmed that @chenghsinw pointed out a good point, closing for fixing.

docs/proposals/2019-09-30 - Content by reference in Link Lists.md
```diff
@@ -0,0 +1,64 @@
...
+```
+## Questions and Concerns
+**So is the idea that an `interstitial_link` now can link to a url, a story, video or gallery? But only one of those would ever be present?**
+Yes and Yes
+
+**Does this mean “show a gallery here” or “show a link to a gallery here?**
+This means show a link to a gallery here.
+
+**What about the potential performance impact if a content producer adds 20 story reference `interstitial_link`s in a Link List?**
+Link Lists are capped at 10 content elements.
```
michaelholm: 1. Does capping the list at 10 items raise any concerns over customers soon wanting 12-15 items, etc.
2. I'm a little confused about the concept here - is the idea to add a link to content rather than embedding the content in the story?

jpblair:
I think WebSked currently ignores content elements of type `interstitial_link`, so this shouldn't affect us.

docs/proposals/2019-09-30 - Content by reference in Link Lists.md
```diff
@@ -0,0 +1,64 @@
+## Problem
+Currently, editors can only add URLs in `interstitial_link` to Link Lists. This presents a few issues:
+1. it's not possible to use a Link List until all of the desired content is published
+2. it's not as straightforward to fetch additional information about the story with just the URL
+
+## Proposal
+We propose to add additional traits, `Story`, `Video`, and `Gallery`, to `interstitial_link`, so we can add more types of elements by reference to link lists in addition of `image`.
```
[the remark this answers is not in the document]
taiweituan: >How does this solve the actual problem? 
According to the Problem, this will enable user to add either of story, video, or gallery to Link List while either of story, video, or gallery is still working in progress/unpublished by using reference instead of the URLs, which is generated after it is published. 
>What happens if the Story or Video or Gallery isn't published?
User should still be able to add them into Link List.

docs/proposals/2019-09-30 - Content by reference in Link Lists.md
```diff
@@ -0,0 +1,64 @@
...
+    }
+  },
+}
+```
+## Questions and Concerns
+**So is the idea that an `interstitial_link` now can link to a url, a story, video or gallery? But only one of those would ever be present?**
+Yes and Yes
+
+**Does this mean “show a gallery here” or “show a link to a gallery here?**
+This means show a link to a gallery here.
```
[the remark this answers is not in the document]
taiweituan: User is provided with required input field to enter **Link Title** for each interstitial links.

docs/proposals/2019-09-30 - Content by reference in Link Lists.md
taiweituan: >Does capping the list at 10 items raise any concerns over customers soon wanting 12-15 items, etc.
It's under discussion, currently the idea is set to 100 or fewer.

>I'm a little confused about the concept here - is the idea to add a link to content rather than embedding the content in the story?
Yes. To be precise, the idea is to add a link (or non-link content, in another PR) to a Link List instead of embedding the content in the story.

taiweituan: >Does capping the list at 10 items raise any concerns over customers soon wanting 12-15 items, etc.

It is an ongoing discussion, currently the idea is set to 100 or fewer. 
>I'm a little confused about the concept here - is the idea to add a link to content rather than embedding the content in the story?

Yes, the idea is adding content(video, story, gallery) to link list container (while allowing them to be added by `promo_reference`) rather than embedding the content in the story individually.

taiweituan:
> Based on the use case presented, I would strongly prefer `promo_reference`s be used instead of "real" references.
Proposal updated, thanks!

PR #221, opened 2019-10-22
Proposal for Text Only Elements in Link Lists

nwdavis:
https://arcpublishing.atlassian.net/browse/COMP-3465

### DIFF
--- a/docs/proposals/2019-10-22_Linkless_Link_Lists.md
+++ b/docs/proposals/2019-10-22_Linkless_Link_Lists.md
@@ -8,7 +8,7 @@ Some clients like the content-element-group-like nature of Link Lists but have u
 
 Make the URL in a link list entry optional, or, to put it another way, allow link lists to support text elements in addition to interstitial links.
 
-# Example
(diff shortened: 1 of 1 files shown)


nwdavis:
> If you have a list of mixed links and text, what makes a link list with some non-links preferable to a normal list with some links?

It's my understanding that customers want a content block behavior, which doesn't currently exist with list elements in Composer. It's likely that they would create link lists with only text items as well, not necessarily mixed.

anscho:
> It's likely that they would create link lists with only text items as well, not necessarily mixed.

If this is really just for styling reasons, then this seems like a counter-intuitive use of a "link list" item, which is described in the specification:

> A list of links to related or external content that may be embedded in a story as a unit. Can be used for curated re-circulation, or simple 'More Information' boxes. Each item in the list must be an interstitial link.

Why not create a new type that is a list of any type of content?

jpblair:
To my knowledge, WebSked ignores content elements of type `link_list`, so would not be affected by this change.

PR #222, opened 2019-10-23
Geo Restrictions Association Proposal: CA-1911

jodieburnett:
Ticket: https://arcpublishing.atlassian.net/browse/CA-1911

A proposal for adding geo-restriction ids to the `content_restrictions` trait.

### DIFF
--- a/docs/proposals/2019-10-22-Geo-Restriction-Ids.md
+++ b/docs/proposals/2019-10-22-Geo-Restriction-Ids.md
@@ -1,0 +1,38 @@
+# Adding Geo-Restriction Information to the content_restrictions Trait
+
+# Problem
+
+Certain clients need the ability to restrict content based on a user's geographical location, and to decide the amount of time that a piece of content is allowed to be publically viewed. To support this, documents must be associated with restrictions in some way.
+
+# Proposal
+
+We propose extending the `content_restrictions` trait to include a `geo` field, which maps to an object that contains the restriction information the document should be associated with. For the moment, only videos should be allowed to use the `geo` field.
+
+## restrictions
+
+The `geo` object should have a `restrictions` field. This field would map to an array of restriction objects, each of which contains a `restriction_id`. The `restrictions` array should be limited to a size of 1 for now.
+
+## Example
+```
+{
+  "type": "video",
+  "version": "0.10.4",
+  "content_restrictions": {
+    "geo": {
+      "restrictions": [{
+        "restriction_id": "abcdefghijklmnopqrstuvwxyz"
+      }]
+    }
+  }
+}
+```
+
+# Concerns
+
+## Why not store the document/restriction associations in PubStack, or in a dedicated new service?
+
+While this would eliminate the need for a later migration, it isn't feasible under the current time constraints.
+
+# Implementation
+
+The Core Services team will implement this if the proposal is accepted. The ticket already exists in the Content API project: [CA-1917](https://arcpublishing.atlassian.net/browse/CA-1917)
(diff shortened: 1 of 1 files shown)


RyanGladstone:
`The restrictions array should be limited to a size of 1 for now.`
Is the idea to lift this cap in the future?

gengel:
@RyanGladstone I think the idea is to leave ourselves the option of enabling it later in a non-breaking way.

jodieburnett:
Putting an in-progress label on this. This proposal may change after some feedback from @rgfindl

docs/proposals/2019-10-22-Geo-Restriction-Ids.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Example
+```
+{
+  "type": "video",
+  "version": "0.10.4",
+  "content_restrictions": {
+    "geo": {
+      "restrictions": [{
+        "restriction_id": "abcdefghijklmnopqrstuvwxyz"
```
chenghsinw: Just curious what's the real geo restriction object should look like?

jodieburnett: The real object would look like that, for now, with just a `restriction_id` field. It's an object so that we can leave it open for adding other fields in the future, if necessary.

anscho: I think @chenghsinw means the restriction definition itself, which we aren't defining here (this is just associating a document with a restriction defined elsewhere).

chadmoone: Are `restriction_id`s arbitrary?

chadmoone: Ok, then yeah, if this ID is a reference to another object, should that not be defined here?

jodieburnett: From what I understand, this reference won't be denormalized at any point, so it won't be part of the ANS document and doesn't need to be defined here.

jodieburnett: @gengel correct me if I'm wrong, or if that's not quite the right justification.

gengel: From the ANS perspective, the ids are an arbitrary string, similar to distributor ids.

The Geo-Restrictions have their own API and spec -- that itself is not ready yet, but is somewhat outside the scope of this PR.

docs/proposals/2019-10-22-Geo-Restriction-Ids.md
```diff
@@ -0,0 +1,38 @@
...
+
+Certain clients need the ability to restrict content based on a user's geographical location, and to decide the amount of time that a piece of content is allowed to be publically viewed. To support this, documents must be associated with restrictions in some way.
+
+# Proposal
+
+We propose extending the `content_restrictions` trait to include a `geo` field, which maps to an object that contains the restriction information the document should be associated with. For the moment, only videos should be allowed to use the `geo` field.
+
+## restrictions
+
+The `geo` object should have a `restrictions` field. This field would map to an array of restriction objects, each of which contains a `restriction_id`. The `restrictions` array should be limited to a size of 1 for now.
```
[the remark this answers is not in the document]
anscho: Hi @design1online. I think this comment addresses the question:
https://github.com/washingtonpost/ans-schema/pull/222#discussion_r338777859

If you have questions about how our project will implement this spec, I'd be happy to chat about it on Slack.

docs/proposals/2019-10-22-Geo-Restriction-Ids.md
yury-sannikov: @chenghsinw I think we need to figure it out in this ticket https://arcpublishing.atlassian.net/browse/AH-826

PR #227, opened 2020-02-18
Proposal - Planned & Actual Character Count

nwdavis:


meg-delaunay:
Seems simple and rational to me. I'll speak for @gengel and say this is 👍.

meg-delaunay:
> I would recommend we also include the charset encoding if we're going to do a number of characters. Different encodings use different numbers of bits per characters and different values to represent different characters, this is especially important with Chinese characters.

I think this is a really good idea.

docs/proposals/2019-02-20-Planned_Actual_Char_Count.md
```diff
@@ -0,0 +1,38 @@
...
+    "properties": {
+      "character_count_planned": {
+        "description": "The anticipated number of characters in the story.",
+        "type": "integer"
+      },
+      "character_count_actual": {
+        "description": "Current number of characters.",
+        "type": "integer"
+      }
+    }
```
design1online: "character_encoding": {
        "description": "The encoding used for counting these characters.",
        "type": "string"
      }

PR #229, opened 2020-05-26
Adds documentation for Draft API handling of ANS fields (PSTK-1054)

jodieburnett:
View the file here: https://github.com/washingtonpost/ans-schema/blob/PSTK-1054/docs/2020-05-26-Draft-API-and-ANS.md

This is a document that will be linked to from the Draft API OpenAPI spec. It describes the fields within ANS blobs that are blocked, stripped, ignored, etc when interacting with Draft API. Once this is reviewed and merged, I'll open a PR in the arcpub repo to add this document's link to the OpenAPI spec.

This should be reviewed by arcpub folks, not the usual ANS reviewers. I just put this file here because clients will be able to view it, and it makes sense since it's all about ANS! :)

One additional comment: We have a ticket open to address the `additional_properties.has_published_copy` field. I've left out any mention of that in this document, since we don't know yet what we plan to do about that field. Let me know if you think we should address it in this doc, or wait until we know what we plan to do.


jodieburnett:
Closing this in favor of a PR in the arcpub repo.

PR #230, opened 2020-07-29
Proposal for Document Contributors field

meg-delaunay:
Proposal for adding a Document Contributors (`contributors`) field to each document type in order to track who created (and in the future, contributed at all) to a document in Arc.

### DIFF
--- a/docs/proposals/2020-07-27-Document_Contributors.md
+++ b/docs/proposals/2020-07-27-Document_Contributors.md
@@ -1,0 +1,50 @@
+
+# Problem
+
+Throughout Arc, there is no canonical way to track who has created or contributed to a particular document. While each Revision of a Story has a user_id associated, with the introduction of the Publishing Platform, there are contributions to a document that can be made without creating a Revision (ie Circulations). Plus, to get the creator, you'd need a way to easily go get the first ever Revision, which can sometimes involve paginating through hundreds of Revisions.
+
+Authors exist in Arc, but pose a problem when content is created by someone who is not really an Author in Arc (ie Admin users) OR created by someone other than the person that should appear on the final reader-facing page (ie ghost writers).
+
+An upcoming use case of Arc Broadcast is a homepage view, which will show app users a feed of content created within their organization's Broadcast app. Ideally, this homepage would be able to show content created in Broadcast by the user themselves, in order to make it easy for the user to jump into their own content while on the go.
+
+# Proposal
+
+Add a `contributors` trait to the Story, Video, Image, Gallery, and Content Types that will hold key information about the person(s) who created and then later edited the Document.
+
+For this MVP, a limit of length 1 on the array is acceptable, holding just the information on the `creator`. However, as we move forward with this feature in Broadcast, there may be a desire to hold a list of all contributors to the Document, which is why it is an Array.
+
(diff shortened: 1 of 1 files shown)


gengel:
Great idea and proposal! I do have some questions.

1) In the future state where this field is adopted, what is the difference between `credits`, `contributors.created_by` and `revision.user_id`? When should each of these be set, and by whom? From the proposal, I have a general sense of it, but there's a lot of overlap.

2) Is `contributors.created_by` intended to be something editorial users can overwrite / control? Is it something that's generated automatically as users interact with a document? Some combination of the two? Given the use case provided, it seems like either could work, but I do think a semantic definition would clarify this.

meg-delaunay:
Here's my general idea of the difference between the three that we should uphold:

- `credits` - The reader-facing Author of the Document; should be an actual Arc `author`; should be used to be displayed to readers 
- `revision.user_id` - The creator of that particular Revision (which I believe applies only to Stories?)
- `contributors.created_by` - The creator of the Document; ie, the `revision.user_id` of the first every Revision 
- (in a future state) `contributors.updated_by` - A list of individuals who have touched the Document in some way, whether creating a new Revision, Circulating, or Publishing (creating a new Published Revision)

The big difference between `credits` and the other fields is that `credits` is reader-facing and should reference an Arc Author (I'm not sure if this is enforced). It also is completely manually set by the User in the Editorial App or via the API. 

`user_id` and `contributors` on the other hand, are purely for editorial uses, and should be set automatically via the apps. I know for sure that Composer is setting `revision.user_id` automatically and passes that to Draft API. I'd expect `contributors.created_by` to work in the same way. 

The question then becomes "why do we need `contributors` when we have `revision.user_id,` which is centered around having an easy way on a Document to get that first creator, without needing to paginate through any number of revisions. 

If there's a better way to represent those differences, I'm all for it.

gengel:
Got it. So at any given point in time, `contributors.created_by` should be the same as `revision.user_id` for the first revision (if it exists) and `contributors.updated_by` should be the same as the aggregated values of `revision.user_id` across all revisions?

meg-delaunay:
Exactly. This just puts those values in a place that can be easily searched and accessed via Content API instead of baked into old Revisions.

RyanGladstone:
I debated with myself whether Broadcast should just use credit (and perform a search by name, like Composer offers*) but I think this will be a net benefit to the platform, particularly when the full contributor list is captured and indexed. This property will open up some interesting possibilities on the Home dashboard and with notifications. 

* example query by credit: https://staging.arcpublishing.com/composer/api/v1/search/?website=raycom&sort=created_date:desc&from=0&size=20&body=%7B%22query%22%3A%7B%22bool%22%3A%7B%22must%22%3A%5B%7B%22term%22%3A%7B%22type%22%3A%22story%22%7D%7D%2C%7B%22match_phrase_prefix%22%3A%7B%22credits.by.name%22%3A%22Ryan%20Gladstone%22%7D%7D%2C%7B%22term%22%3A%7B%22revision.published%22%3A%22false%22%7D%7D%2C%7B%22terms%22%3A%7B%22workflow.status_code%22%3A%5B6%2C8%2C7%2C2%2C9%5D%7D%7D%5D%7D%7D%7D

rgfindl (approved):

jimas14 (approved):

PR #231, opened 2020-08-14
PSTK-1362: Collection Query Backfill Proposal

mfattoru:


docs/proposals/2020-08-12-Collection_Query_Backfill.md
```diff
@@ -0,0 +1,69 @@
...
+Backfilling Collections with query results is a complex process that requires implementation across multiple systems.
+This document proposes a first solution to achieve the near-term deliverable.
+
+## Proposal
+
+In order to limit scope on both the rendering and publishing sides, we only allow specifying a single section or tag. The Collections API would be updated to support this via the API and presumably UI as well.
+
+### Example
+
+We would update the ANS spec to allow specifying a single website or tag:
```
tbrick855, line 14: I think you mean section not website

docs/proposals/2020-08-12-Collection_Query_Backfill.md
```diff
@@ -0,0 +1,69 @@
...
+
+```JSON
+{
+  "type": "collection",
+  "dynamic_items": {
+    "type": "sections"
```
tbrick855, line 20: You say single section but the type is "sections" and it uses an array

mfattoru: explained the use of plural and array

docs/proposals/2020-08-12-Collection_Query_Backfill.md
```diff
@@ -0,0 +1,72 @@
...
+
+`https://api.washpost.arcpublishing.com/content/v4/search/published?website=washpost&q=type:story+AND+taxonomy.sites._id:%22/politics%22&sort=display_date:desc&size=10`
+
+#### Tag-based example query:
+
+`https://api.washpost.arcpublishing.com/content/v4/search/published?website=washpost&q=type:story+AND+taxonomy.tags.slug:obama&sort=display_date:desc&size=10`
+
+### Are there limits on the number of dynamically-filled items?
+
+I suggest limiting it to the same as Content API’s general limit, which would be a total of 100 items.
```
brendaneamon, line 64: I imagine that customers might expect any limit to respect the collection's `cap`, which currently defaults to its maximum value of `500`. If an organization were to have a collection with a cap of 50 and 20 hand-curated content items, they would probably expect a maximum of 30 additional stories to be returned by the query backfill.

RyanGladstone:
```How are dynamically filled items marked inside the collection?

We could add is_dynamic_item under additional_properties, but I’m open to alternatives.```

Looks like this is already merged but I have a question and a comment:

1. Is there a client need to differentiate between manually curated and backfilled items within a collection? I certainly think it's possible but haven't heard any requests in this area and can't immediately think of a use case

2. I think it is an Arc goal to move platform-supported fields out of additional properties wherever possible. Perhaps there's not a mandate on this front.

docs/proposals/2020-08-12-Collection_Query_Backfill.md
RyanGladstone: Hm, but how does the Collection cap interact with the result set size in the Content API request?

RyanGladstone: Oh, really? I read it how you originally read it

RyanGladstone: `https://api.washpost.arcpublishing.com/content/v4/search/published?website=washpost&q=type:story+AND+taxonomy.sites._id:%22/politics%22&sort=display_date:desc&size=10`

Given a Collection with five manually curated items and a cap of 20, my expectation is that the Content API would return 10 stories, including five backfilled from the Politics section.

brendaneamon: I thought

> Oh, really? I read it how you originally read it

I thought so too until I saw the implementation, so this probably merits clarification.

mfattoru: yes, the first implementation will limit the number of tags sections to one, which will be restricted to the application layer. A second iteration will expand the same number to 100

PR #237, opened 2021-02-03
PPSE-912: flipped author and site schema in operations schema

manginav:
### Description 
Schemas are flipped between author and site, so flipped them back


mfattoru:
If this error was introduced in 0.10.0, should this fix all versions from 0.10.0 and up?

manginav:
> If this error was introduced in 0.10.0, should this fix all versions from 0.10.0 and up?

Hi @mfattoru I am not sure, but this is same across all the versions. But since they have pointed for 0.10.0 I have made the change for all the version up. In any case it is typo so it would be good to fix it is what I thought

mfattoru:
I also just merged in 0.10.8 that needs to be updated. Also, We might want to wait for someone else comments, as I'm not sure how issues on older releases should be tackled

manginav:
> I also just merged in 0.10.8 that needs to be updated. Also, We might want to wait for someone else comments, as I'm not sure how issues on older releases should be tackled

added changes for 0.10.8 @mfattoru  👍

manginav:
Hi @mfattoru  I have changed this to only for versions 0.10.0 and 0.10.8

manginav:
done with your suggestion @mfattoru

mfattoru (approved):

PR #239, opened 2021-03-23
PPSE-1454: added return after callback to avoid code execution if there is an error or if it is finished

paoamaralwz:
added return after callback to avoid code execution if there is an error or if it is finished

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@
 ### 1.10.7.1 2021/3/30
 
 * Adds a return when it finish reading files to retrive ANS versions
+
 ### 1.9.1 2019/2/1
 
 * Adds new ANS version to validator logic
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 
 ### 1.10.7.1 2021/3/30
 
-* Adds a return when it finish reading files to retrive ANS versions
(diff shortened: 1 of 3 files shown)


calexicoz (approved):
@paoamaralwz can you update the version number in `package.json` to `1.10.7.1`? This will be needed for the release. We should also include some notes in the `RELEASE_NOTES` file, which has not been updated properly.

Thanks!

paoamaralwz:
> @paoamaralwz can you update the version number in `package.json` to `1.10.7.1`? This will be needed for the release. We should also include some notes in the `RELEASE_NOTES` file, which has not been updated properly.
> 
> Thanks!

ok, I will update the version and update RELEASE_NOTES

RELEASE_NOTES.md
```diff
@@ -1,5 +1,9 @@
 # ANS Release Notes
 
+### 1.10.7.1 2021/3/30
+
+* Adds a return when it finish reading files to retrive ANS versions
```
calexicoz, line 5:
```suggestion
* Prevent the callback function from being called twice when errors occur during the loading of schemas
```

calexicoz (commented):
Hey @paoamaralwz,

Sorry I just noticed this but the issue appears for both `loadAllSchemas` and `loadSchema`. Can you fix it for the other one too?

Thanks.

calexicoz (approved):

PR #258, opened 2022-06-28
Auth and SEO Filename Proposals

jimas14:
These two fields are being proposed by Photo Center as image traits, to be used in [Resizer V2](https://github.com/WPMedia/arc-photo-resizer).

### DIFF
--- a/docs/proposals/2022-06-27_Auth_Proposal.md
+++ b/docs/proposals/2022-06-27_Auth_Proposal.md
@@ -2,7 +2,7 @@
 
 # Problem
 
-The [Signing Service](https://github.com/WPMedia/signing-service) was created to provide Photo Center & PageBuilder HMAC tokens which allow the use of images on the frontend. Without a field to store these tokens in, users would have to call the Signing Service for each image they want to display.
+The [Signing Service](https://github.com/WPMedia/signing-service) was created to provide Photo Center & PageBuilder HMAC tokens which allow the use of images on the frontend. Without a field to store these tokens in, users would have to call the Signing Service for each image they want to display using [Resizer V2](https://github.com/WPMedia/arc-photo-resizer).
 
 # Proposal
 


docs/proposals/2022-06-27_SEO_Filename_proposal.md
```diff
@@ -0,0 +1,44 @@
+# SEO Filename Proposal
+
+# Problem
+
+Image resizer links in Arc are pretty ugly. The only reference to an image filename is the Arc ID, like `ABCDEFGHIJKLMNOPQRSTUVWXYZ.jpg`. If clients want to improve the SEO scoring of their image links in Google, they need to be able to make that filename more realistic.
+
+# Proposal
+
+Like Canonical URLs do for gallery leaf pages, we want to introduce a string field SEO Filename. This field will be used by PageBuilder and others when generating Resizer V2 URLs by appending it before the asset key, like `/resizer/v2/this-is-an-seo-filename-ABCDEFGHIJKLMNOPQRSTUVWXYZ.jpg`.
```
Greymalkin, line 9: lgtm

docs/proposals/2022-06-27_Auth_Proposal.md
```diff
@@ -0,0 +1,52 @@
...
+    "type": "image",
+    "version": "0.10.9"
+}
+```
+
+# Concerns
+
+> Aren't you exposing a token in ANS?
+
+Yes but it is the signed token, so there is no risk of exposing the secret key. Plus, it's only valid for that single Arc object.
```
Greymalkin, line 43: I want to step back and let others with more paranoid feelings about security speak up on this one.  I'm okay with this, but I'm interested to know if anyone else is not.

Greymalkin (commented):
I'm fully on board with the resizer name proposal.  I'd like to wait and hear someone else's opinion on the auth token proposal.

docs/proposals/2022-06-27_SEO_Filename_proposal.md
```diff
@@ -0,0 +1,44 @@
...
+
+```
+{
+    "_id": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
+    "additional_properties": {
+        ...
+    },
+    "caption": "This is a caption.",
+    ...
+    "seo_filename": "this-is-an-seo-filename",
```
darful, line 25: Is the plan to add this to just the Image Content-Type?

darful (changes requested):
The SEO proposal LGTM.

The Auth proposal could use a pass fleshing out the solution in more detail, please see my in-line comments.

docs/proposals/2022-06-27_Auth_Proposal.md
```diff
@@ -0,0 +1,52 @@
+# Auth Proposal
+
+# Problem
+
+The Signing Service produces SHA256-signed HMAC tokens that are used to validate requests for Arc objects on the front-end. Those front-ends must make a call to Signing Service for each object that it wants to display.
```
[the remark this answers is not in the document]
jimas14: @darful I just updated the doc, could take another look when you get a chance?

docs/proposals/2022-06-27_SEO_Filename_proposal.md
jimas14: Yup!

docs/proposals/2022-06-27_Auth_Proposal.md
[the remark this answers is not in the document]
darful: Ah, the context that we are solving a perceived problem helps a lot.

docs/proposals/2022-06-27_Auth_Proposal.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Concerns
+
+> Aren't you exposing a token in ANS?
+
+It is just the signed token, so there is no risk of exposing the secret key, and the token is only valid for that single image.
+
+> What prevents someone from reading the token from ANS embedded on a website and using it?
+
+We decided to accept this risk now that we aren't worried about the performance of Thumbor in ECS. Akamai is much more durable and has protections in place for DDOS & other exploits.
```
darful, line 53: I see.  This is security through obfuscation which is marginally better than no security at all.

For token based security like this, being able to revoke a token is typically a requirement as eventually a bad actor will figure out our obfuscation pattern.  How does token revocation and replacement work for this system?  Simply create a new versioned token and deactivate the compromised one?

My concern here is that once a bad actor understands our obfuscation pattern then this auth system is effectively broken.  Most client-side auth systems suffer from this drawback and this drawback is hedged against by using secure granular tokens with limited access rights and a token revocation/replacement process.  

Can you document the token revocation/replacement process in this proposal?
Can you document the limitations of this auth system?

Lastly, we should get AppSec on this PR once the proposal is finalized.

darful (changes requested):
Please see my in-line comments.

docs/proposals/2022-06-27_Auth_Proposal.md
darful: @ibaker thanks for the in-depth explanation of the auth system this ANS change is interacting with as the context helps explain most of the why of this PR.

darful (approved):

PR #277, opened 2023-10-20
Add host_talent and contributors fields to credits trait

edziegle:
5 years ago, a proposal was accepted to add `host_talent` and `contributors` to the `credits` trait for video content. However, this was never formally implemented. This PR intends to finally correct that.


edziegle:
See #191

jimas14 (approved):

edziegle:
> Howdy.
> 
> This looks ok, but every change to the ANS schema requires changing the ANS version. Can you do this:
> 
> > https://github.com/washingtonpost/ans-schema#creating-a-new-ans-version
> 
> Once the new version has been created, there is an internal release process to go through to make it available to use.
> 
> Thanks!

Should I go ahead and move this into the current dev version, or just make a 0.10.12?

edziegle:
Going to close this for now, but will revisit it later.

PR #280, opened 2024-03-01
Intelligence Team - Proposal for new fields for content analysis

pardoc3:


cselvaraj (approved):
This looks good to me!

I have a comment, which I think is due to my ignorance of the existing Clavis fields' usage.  Before I dive into that, just a quick comment:

* `src/main/resources/schema/ans/0.10.10/traits/trait_taxonomy.json` schema should also be listed among the files that will be changed by this proposal

These new fields seem similar to the Clavis fields, but have some slight differences.  I was given some context about the need, and was told that the existing fields were not suitable for the purposes here.  I'm good with these new fields being added, given that situation.  Furthermore, the old fields are labeled with the Clavis name, so I'm not concerned with there being confusion about the fields' meaning.

However, I would just ask whether these new fields are flexible enough to be generally useful or if they just fit the current need?  I'm certainly not qualified to answer that question.  I'd like to avoid the situation where an ever increasing number of fields are added to the `taxonomy` object.  Assuming Clavis' purpose is similar to your own, the ideal situation would be that Clavis could use these new fields to accomplish their goals as well.  If I'm wrong about the similar purpose though, my concern can be ignored.

Ultimately, if you feel confident that others could use these fields for similar ends or there is no more generic way to structure these fields, I see no reason to bar their addition.

jecerovac2:
> This looks good to me!
> 
> I have a comment, which I think is due to my ignorance of the existing Clavis fields' usage. Before I dive into that, just a quick comment:
> 
> * `src/main/resources/schema/ans/0.10.10/traits/trait_taxonomy.json` schema should also be listed among the files that will be changed by this proposal
> 
> These new fields seem similar to the Clavis fields, but have some slight differences. I was given some context about the need, and was told that the existing fields were not suitable for the purposes here. I'm good with these new fields being added, given that situation. Furthermore, the old fields are labeled with the Clavis name, so I'm not concerned with there being confusion about the fields' meaning.
> 
> However, I would just ask whether these new fields are flexible enough to be generally useful or if they just fit the current need? I'm certainly not qualified to answer that question. I'd like to avoid the situation where an ever increasing number of fields are added to the `taxonomy` object. Assuming Clavis' purpose is similar to your own, the ideal situation would be that Clavis could use these new fields to accomplish their goals as well. If I'm wrong about the similar purpose though, my concern can be ignored.
> 
> Ultimately, if you feel confident that others could use these fields for similar ends or there is no more generic way to structure these fields, I see no reason to bar their addition.

These new fields are generic and can be leveraged by any tagging solution a client wants to put in place.  The three fields requested are the minimum must-have fields; I did request the addition of two nice-to-have fields (custom categories and custom entities) that will future-proof this so that it is flexible enough to meet any client's needs without having to do any more updates later.

docs/proposals/2024-03-01_new_fields_for_content_analysis.md
```diff
@@ -0,0 +1,253 @@
...
+        "$ref": ".../content_topic.json"
+      },
+      "description": "A list of topics. Topics are the subjects that the content is about."
+    },
+"entities": {
+      "type": "array",
+      "items": {
+        "$ref": ".../entity.json"
+      },
+      "description": "A list of entities. Entities are proper nouns, like people, places, and organizations."
```
mayela, line 222: I think below this line you can add the 2 fields: custom_categories and custom_entities

jecerovac2 (commented):
Looks good to me.

ramirezabraham:
Looks good to me as well.

cselvaraj:
Additional changes look good to me.

cselvaraj:
@pardoc3, if there are no further changes to the proposal, feel free to merge.  We'll just need another PR with the schema changes and we can work to get these changes released.

cselvaraj:
Oops.  I've been pointed to the other PR:

https://github.com/washingtonpost/ans-schema/pull/281

I'm merging this.

PR #282, opened 2024-04-25
PPA-825: fix ans-schema version for fixing upvert failures

manginav:
### Ticket 
https://arcpublishing.atlassian.net/browse/PPA-825

### ANS _SCHEMA upvert failure
- upvert is failing because nested objects are checked for version. 
- from_version.compare(to_version) cannot parse versions like 1 
- And upvert can also fail if those type checks are for versions below 0.5.0, because versions are only present from 0.5.0 in ans-schemas, 
- so type checks are only necessary for those types which have and follow semantic version types like https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.10.3/audio.json#L17
- for now we are hard coding in `version_upverter_types` we need to update this to be dynamic from schemas.


tests/transform-tests.js
```diff
@@ -690,4 +692,45 @@ describe("Transformations: ", function() {
...
+              type: "custom_embed",
+            },
+          ],
+        };
+        var result = transforms.upvert(original_ans, LAST_VERSION);
+        assert.ok(result instanceof Error === false);
+        if (Version.is_semantic_version(version)) {
+          const expected = _.cloneDeep(original_ans);
+          expected.version = LAST_VERSION;
+          expected.content_elements[0].embed.config.editorState.root.version =
```
cselvaraj, line 725: This version shouldn't be modified.  It's the version of the config for the custom embed.  That's effectively an internal value for the custom embed.  We need to ensure this number is never modified by the upvert process.

lib/transforms.js
```diff
@@ -41,6 +41,10 @@ var doUpvert = function doUpvert(ans, version_to) {
     return ans;
   }
 
+  if(!version.is_semantic_version(version_from.str()) || !version.is_semantic_version(version_to.str())){
+    return ans
+  }
```
manginav, line 46: without this we will get a uncaught exception.

lib/transform_utils.js
```diff
@@ -32,8 +33,9 @@ var version_incrementer = function(new_version) {
           });
         }
       });
+      
 
-      if (_.has(ans, 'version') ||
+      if (_.has(ans, 'version') && Version.is_semantic_version(ans.version) &&
```
manginav, line 38: This saves from not touching versions which are not in the top_types and which are not semantic else inner versions during recursive transform are change to latest ans version, example Version: 1 of type `children` in custom_embed will be changed to Version: 0.10.10

PR #284, opened 2024-04-29
PPA-825: fix ans-schema version for fixing upvert failures

cselvaraj:
(Remade to rebase https://github.com/washingtonpost/ans-schema/pull/282)

### Ticket 
https://arcpublishing.atlassian.net/browse/PPA-825

### ANS _SCHEMA upvert failure
- upvert is failing because nested objects are checked for version. 
- from_version.compare(to_version) cannot parse versions like 1 
- And upvert can also fail if those type checks are for versions below 0.5.0, because versions are only present from 0.5.0 in ans-schemas, 
- so type checks are only necessary for those types which have and follow semantic version types like https://github.com/washingtonpost/ans-schema/blob/master/src/main/resources/schema/ans/0.10.3/audio.json#L17
- for now we are hard coding in `version_upverter_types` we need to update this to be dynamic from schemas.


cselvaraj (commented):
LGTM!

manginav (approved):

PR #288, opened 2025-05-16
ACMS-1317: Adding pattern to validate URL

sneh-joshi:
This pull request introduces a new pattern validation for external canonical URLs in the ANS schema and updates the version to `1.12.5`. The key changes include adding a regex pattern for external canonical URLs, ensuring they follow a valid URL structure, and updating the release notes and package version accordingly.

### Schema Enhancements:
* Added a regex pattern to validate the structure of external canonical URLs in `trait_canonical_url_external.json` for both schema versions `0.10.11` and `0.10.12`. This ensures that the `canonical_url` field adheres to a valid URL format. [[1]](diffhunk://#diff-4f23e243d403e39090b19b3dc58889e561b19a1bcb561ea0777a8d080811cc86L6-R7) [[2]](diffhunk://#diff-c1cf9cb47b238dd0c15334a16f6fb43a5d779418e925eaabfb1a1017e8613fd0L6-R7)

### Versioning and Documentation:
* Updated `package.json` to increment the version from `1.12.4` to `1.12.5`.
* Added a new entry in `RELEASE_NOTES.md` for version `1.12.5`, documenting the addition of the pattern for external canonical URL support.

### DIFF
--- a/src/main/resources/schema/ans/0.10.11/traits/trait_canonical_url_external.json
+++ b/src/main/resources/schema/ans/0.10.11/traits/trait_canonical_url_external.json
@@ -4,5 +4,5 @@
   "title": "Canonical External URL",
   "description": "The absolute URL of the document on website which will override any relative URL specified by the `canonical_url` field.",
   "type": "string",
-  "pattern": "^https?:\\/\\/([a-zA-Z0-9]([a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9])?\\.)+[a-zA-Z]{2,}(\\/[a-zA-Z0-9\\-._~:/%]+)*$"
+  "pattern": "^https?:\\/\\/[^\\s]*$"
 }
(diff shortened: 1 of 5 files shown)


gengel (approved):

src/main/resources/schema/ans/0.10.11/traits/trait_canonical_url_external.json
```diff
@@ -3,5 +3,6 @@
   "id": "https://raw.githubusercontent.com/washingtonpost/ans-schema/master/src/main/resources/schema/ans/0.10.11/traits/trait_canonical_url_external.json",
   "title": "Canonical External URL",
   "description": "The absolute URL of the document on website which will override any relative URL specified by the `canonical_url` field.",
-  "type": "string"
+  "type": "string",
+  "pattern": "^https?:\\/\\/([a-zA-Z0-9]([a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9])?\\.)+[a-zA-Z]{2,}(\\/[a-zA-Z0-9\\-._~:/%]+)*$"
```
cselvaraj, line 7: I think this forces URL encoding and precludes query strings, since the question mark is omitted, correct?  I think we might see issues with the latter item.  For the former, I'm somewhat conflicted.  Forcing that might be good in many cases.  But, it may be the case that URLs are stored in their unencoded form in other CMSs.  If we take on a client like this, we may not be able to import their canonical URLs without issue.  In the cases where we get this pattern wrong, it would take an ANS version update to fix.

To avoid this situation, it would probably be good to make this pattern more permissive.  Maybe just `^https?://.+$`.  That would avoid the quagmire of making sure we get this pattern just right.  Is there a business case to make this specific that I'm missing?

cselvaraj (approved):
LGTM!

